PR #8, opened 2018-07-19
Pre-compile auspice and run the non-dev server for a better experience

tsibley:
See commit messages.


trvrb:
Good idea to switch to production. This is working for me. Thanks @tsibley.

PR #10, opened 2018-07-30
update auspice wrapper

jameshadfield:
Updates the auspice wrapper to use the correct `npm` script, as local builds are now specified via URLs, not args to the running server.


tsibley:
Merged. Thanks!

PR #11, opened 2018-08-04
Push unique image tags

trvrb:
@tsibley ---

This PR gives every build / image a unique tag. I chose to use a date string for this. This was following advice here: https://blogs.msdn.microsoft.com/stevelasker/2018/03/01/docker-tagging-best-practices-for-tagging-and-versioning-docker-images/. This would allow the cli to refer to a specific image version rather than just latest. 

This was tested on Travis CI and resulted in the tag `2018-08-04-21-46-25` seen here: https://hub.docker.com/r/nextstrain/base/tags/.

I'm not at all wedded to the date string as tag, if you have better idea. Date string did seem offer convenience and to be easily understandable. I don't think we can use a proper version here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,24 +8,29 @@ services:
   - docker
 sudo: required
 before_install:
+  - export DATESTRING=$(date +%F-%H-%M-%S)
   - docker pull nextstrain/base-builder
   - docker pull nextstrain/base
 install:
   - |
-    docker build --build-arg CACHE_DATE="$(date)" \
+    docker build --build-arg CACHE_DATE=$DATESTRING \
       --cache-from nextstrain/base-builder \
       --cache-from nextstrain/base \
-      -t nextstrain/base-builder --target builder .
+      -t nextstrain/base-builder:latest --target builder .
   - |
-    docker build --build-arg CACHE_DATE="$(date)" \
+    docker build --build-arg CACHE_DATE=$DATESTRING \
       --cache-from nextstrain/base-builder \
       --cache-from nextstrain/base \
-      -t nextstrain/base .
+      -t nextstrain/base:latest .
 script:
   - echo "Build finished"
 after_success:
   - if [[ "$TRAVIS_PULL_REQUEST" == "false" && "$TRAVIS_BRANCH" == "master" ]]; then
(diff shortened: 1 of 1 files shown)


tsibley:
I agree we should be tagging images with more than just `latest`.  That was my expectation and intent (see the `--image` option to `nextstrain build`, as evidence).  I thought, however, that we'd use a small collection of what are called "stable tags" in the article you linked to.  Perhaps tags based on the versions of augur and auspice (the two main Nextstrain components in our images).

I can see the possible benefits of tagging every image with a name, but I don't see a ton of cases where it improves matters.  Can you elaborate with the benefits you see?

If we _are_ going to use dates, I'd rather use an accepted ISO 8601 form like `20180806T175423Z` (and guarantee that they're always in UTC). e.g. `date --utc +%Y%m%dT%H%M%SZ`.  We should also consider if 1s is enough of a resolution to avoid race conditions on Travis. We may want to add milliseconds to the timestampe (with `.%3N`).

trvrb:
I had thought that tagging based on augur and auspice versions wouldn't be enough and that you'd frequently get updates to other components (IQTREE, etc..). Would you just not tag the image in this case? Otherwise, I don't really have a preference if you'd like to implement something else here.

tsibley:
> I had thought that tagging based on augur and auspice versions wouldn't be enough and that you'd frequently get updates to other components (IQTREE, etc..). Would you just not tag the image in this case?

Well, we're using specific versions of IQTREE, FastTree, mafft, etc in the image. They won't update unless we manually change the Dockerfile. Once an image was tagged with an augur/auspice version, those versions wouldn't be tagged again. I'm not sure this is a good idea or not, but it seems more interpretable/actionable than dates. Maybe not?

I guess I'm still not clear on the benefits of tagging every image. Can you expand on why you'd like it? What problem is it the solution to?

trvrb:
Sorry. I don't need every image to have a unique tag. I just didn't see another way to do things. If you have a strategy for tagging you'd prefer to use, I'm definitely game for it.

tsibley:
I think I'm missing something. Maybe we can chat about this in person this afternoon?

trvrb:
This was implemented in c6b04ca896370e42f2ce0d42744045a4621c421a.

PR #12, opened 2018-08-04
Build image using auspice release branch

trvrb:
@tsibley ---

We want a stable call to `nextstrain view`. Better to have the Docker image `latest` tag refer to `release` branch of auspice and only update the Docker image when `release` branch is updated, rather than when `master` is updated.

Currently, auspice Travis CI script properly pings the `docker-base` rebuild only when auspice `release` is pushed: https://github.com/nextstrain/auspice/blob/master/.travis.yml#L11. This change brings `docker-base` in line.

Once we have a `release` branch and proper versioning in augur, I would swap augur to do the same.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -122,7 +122,7 @@ RUN curl -fsSL https://api.github.com/repos/nextstrain/augur/tarball/master \
 # auspice
 WORKDIR /nextstrain/auspice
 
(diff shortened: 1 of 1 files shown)


tsibley:
Yes, I agree. Thanks for changing this. Merged.

BTW, it's good form to line-wrap commit messages to 72 or 80 chars so they display nicely in a terminal instead of running on a single line:

screenshot from 2018-08-06 10-43-26

PR #18, opened 2019-07-11
Upgrade Snakemake to 5.5.2 (the latest stable version)

trvrb:
Checkpoints are a useful newer feature of Snakemake which are only available on 5.4.0 and newer.


trvrb:
I'm going to go ahead and merge this. I don't see any issues in terms of backward compatibility and I need this in place to run builds like checkpointing via `--aws-batch`.

PR #21, opened 2020-03-04
Replace Alpine-based image with a Debian-based image

huddlej:
Alpine Linux uses musl libc instead of glibc which means it cannot use most
pre-built Python wheels for glibc-based Linux distributions [1]. As a result, Python
packages need to be built from source, leading to longer build times and
increased complexity in the Dockerfile for more complex packages. For complex
packages like opencv-python, building from source is especially non-trivial.

This commit ports the existing Alpine-based Dockerfile to a minimal Debian-based
file with pre-installed Python 3 (the official Python Docker image) and removes
dependencies that were originally required to build Python packages from
sources (e.g., cvxopt) where installation from a wheel is now possible.

This commit also removes all explicit references to Python 2 dependencies, as
fauna is compatible enough with Python 3 for downloading to work and sacra is
not actively under development.

This new image has been tested with the zika-tutorial, zika, and seasonal-flu
repositories.

[1] https://github.com/docker-library/docs/issues/904

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,30 +5,24 @@
 # image with tools only needed during the image build.
 
 # First build the temporary image.
-FROM alpine:3.9.6 AS builder
+FROM python:3.7-slim-buster AS builder
 
 # Execute subsequent RUN statements with bash for handy modern shell features.
-RUN apk add --no-cache bash
 SHELL ["/bin/bash", "-c"]
 
 # Add system deps for building
-RUN apk add --no-cache \
+RUN apt-get update && apt-get install -y --no-install-recommends \
         autoconf \
         automake \
-        build-base \
+        build-essential \
         ca-certificates \
         curl \
-        freetype-dev \
         git \
-        gmp-dev \
+        libgmp-dev \
         libpng-dev \
-        linux-headers \
         nodejs \
-        nodejs-npm \
-        perl \
-        python3-dev \
-        suitesparse-dev \
-        xz
+        npm \
+        pkg-config
 
 # Downloading dependencies, these should be pinned to specific versions
 
@@ -61,37 +55,6 @@ RUN curl -fsSL https://github.com/vcftools/vcftools/releases/download/v0.1.16/vc
   | tar xzvpf - --strip-components=2
 RUN ./configure --prefix=$PWD/built && make && make install
 
-# Install Python 3 dependencies
-# These may be upgraded by augur/setup.py,
-# but having them here enables caching
-# Be sure to install numpy ahead of pandas
-
-# cvxopt install is particularly fussy.
-# It is separated out from the rest of the installs to ensures that pip wheels
-# can be used for as much as possible, since using --global-option disables use
-# of wheels.
-RUN CVXOPT_BLAS_LIB=openblas \
-  CVXOPT_LAPACK_LIB=openblas \
-    pip3 install --global-option=build_ext \
-      --global-option="-I/usr/include/suitesparse" \
-      cvxopt==1.1.9
-RUN pip3 install numpy==1.19.4
-RUN pip3 install scipy==1.3.3
-RUN pip3 install pandas==1.1.5
-RUN pip3 install bcbio-gff==0.6.6
-RUN pip3 install biopython==1.76
-RUN pip3 install boto==2.49.0
-RUN pip3 install ipdb==0.10.1
-RUN pip3 install jsonschema==3.0.1
-RUN pip3 install matplotlib==2.2.2
-RUN pip3 install phylo-treetime==0.8.1
-RUN pip3 install requests==2.22.0
-RUN pip3 install rethinkdb==2.4.8
-RUN pip3 install seaborn==0.9.0
-RUN pip3 install snakemake==5.10.0
-RUN pip3 install unidecode==1.1.1
-RUN pip3 install xlrd==1.2.0
-
 # Install envdir, which is used by pathogen builds
 RUN pip3 install envdir==1.0.1
 
@@ -100,6 +63,9 @@ RUN pip3 install awscli==1.18.195
 
 # Install our own CLI so builds can do things like `nextstrain deploy`
 RUN pip3 install nextstrain-cli==2.0.0.post1
+
+# Install Snakemake
+RUN pip3 install snakemake==5.10.0
 
 # Add Nextstrain components
 
@@ -127,41 +93,37 @@ RUN pip3 install --requirement=/nextstrain/fauna/requirements.txt
 # accessible and importable.
 RUN pip3 install --editable /nextstrain/augur
 
+# Install additional "full" dependencies for augur.
+# TODO: these versions should be updated in augur's setup.py to work with the
+# pip install nextstrain-augur[full] approach.
+RUN pip3 install cvxopt==1.2.4
+RUN pip3 install matplotlib==2.2.2
+RUN pip3 install seaborn==0.9.0
+
 # Install Node deps, build Auspice, and link it into the global search path.  A
 # fresh install is only ~40 seconds, so we're not worrying about caching these
 # as we did the Python deps.  Building auspice means we can run it without
 # hot-reloading, which is time-consuming and generally unnecessary in the
 # container image.  Linking is equivalent to an editable Python install and
 # used for the same reasons described above.
-RUN cd /nextstrain/auspice && npm install && npm run build && npm link
-
+RUN cd /nextstrain/auspice && npm update && npm install && npm run build && npm link
 
 # ———————————————————————————————————————————————————————————————————— #
 
-
 # Now build the final image.
-FROM alpine:3.9.6
+FROM python:3.7-slim-buster
 
 # Add system runtime deps
-RUN apk add --no-cache \
+RUN apt-get update && apt-get install -y --no-install-recommends \
         bzip2 \
         ca-certificates \
         curl \
-        bash \
-        freetype \
-        gmp \
         gzip \
-        lapack \
-        libpng \
         nodejs \
         perl \
-        python2 \
-        python3 \
-        suitesparse \
         ruby \
-        tar \
         wget \
-        xz \
+        xz-utils \
         zip unzip
 
 # Configure the prompt for interactive usage
@@ -189,7 +151,7 @@ RUN curl -fsSL https://github.com/neherlab/nextalign/releases/latest/download/ne
 RUN chmod a+rX /usr/local/bin/* /usr/local/libexec/*
 
 # Add installed Python libs
-COPY --from=builder /usr/lib/python3.6/site-packages/ /usr/lib/python3.6/site-packages/
+COPY --from=builder /usr/local/lib/python3.7/site-packages/ /usr/local/lib/python3.7/site-packages/
 
 # Add installed Python scripts that we need.
 #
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -143,6 +97,13 @@ RUN pip3 install --requirement=/nextstrain/fauna/requirements.txt
 # accessible and importable.
 RUN pip3 install --editable /nextstrain/augur
 
+# Install additional "full" dependencies for augur.
+# TODO: these versions should be updated in augur's setup.py to work with the
+# pip install nextstrain-augur[full] approach.
```
tsibley: :+1: for updating setup.py to declare these extras and then using that extra tag here.

Dockerfile
```diff
@@ -151,33 +112,25 @@ RUN pip3 install --editable /nextstrain/augur
 # used for the same reasons described above.
 RUN cd /nextstrain/auspice && npm install && npm run build && npm link
 
+RUN apt-get clean
```
tsibley: Because each directive in Dockerfile creates a new layer, this `apt-get clean` will remove apt's cache files from the layer which it creates, but the layers before it will still contain the files and still must be downloaded with the rest of the layers. The typical solution would be to bundle the `apt-get clean` in the same `RUN` as the `apt-get install`, but the Debian image this is ultimately based on helpfully pre-configures apt to run what's effectively `apt-get clean` after every install:

https://github.com/debuerreotype/debuerreotype/blob/6b722a49935a94a9f718f07616f0863db6267023/scripts/debuerreotype-minimizing-config#L85-L109

Which is all to say, this line isn't working as intended and I think can just be removed. :-)

Dockerfile
```diff
@@ -5,30 +5,26 @@
...
+FROM python:3.7-slim-buster AS builder
 
 # Execute subsequent RUN statements with bash for handy modern shell features.
-RUN apk add --no-cache bash
 SHELL ["/bin/bash", "-c"]
 
 # Add system deps for building
-RUN apk add --no-cache \
+RUN apt-get update
+RUN apt-get install -y --no-install-recommends \
```
tsibley: Common practice is to bundle the `apt-get update && apt-get install …` to reduce the size and number of layers.

Dockerfile
```diff
@@ -151,33 +112,25 @@ RUN pip3 install --editable /nextstrain/augur
...
-        perl \
-        python2 \
-        python3 \
-        suitesparse \
-        ruby \
-        tar \
         wget \
         zip unzip
 
+RUN apt-get clean
```
tsibley: Ditto as above.

Dockerfile
```diff
@@ -151,33 +112,25 @@ RUN pip3 install --editable /nextstrain/augur
...
         gzip \
-        lapack \
-        libpng \
         nodejs \
-        perl \
-        python2 \
-        python3 \
-        suitesparse \
-        ruby \
-        tar \
```
tsibley: I assume you were able to remove Bash, Perl, Ruby, tar, etc because they're already installed?

If they're not already installed, I do think it'd be useful to keep them so that builds can easily incorporate custom Ruby or Perl scripts, for example.

Dockerfile
```diff
@@ -118,8 +74,8 @@ ARG CACHE_DATE
 # Add download helper
 COPY devel/download-repo /devel/
 
-# sacra
-RUN /devel/download-repo https://github.com/nextstrain/sacra master /nextstrain/sacra
+# # sacra
+# RUN /devel/download-repo https://github.com/nextstrain/sacra master /nextstrain/sacra
```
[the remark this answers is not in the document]
huddlej: Same here. I plan to do a final rebase without these lines prior to merge. But it is probably safer to just delete in the future, so I don't forget.

Dockerfile
huddlej: Awesome! Thank you for digging that up from the Python image! I've removed these lines.

Dockerfile
huddlej: bash, perl, and tar were already in the base image, but ruby was not. I've added ruby back to the list of packages installed in the final image.

Dockerfile
tsibley: :-) I was pretty sure the base image did this auto-cleaning, but couldn't find it documented, so I dug into the source quickly to make sure I wasn't making things up.

Dockerfile
tsibley: Great!

tsibley:
@huddlej I really want to merge this and leave Alpine behind, but I'm hesitant because it seems like a large underlying change that could break existing `nextstrain build` uses between updates. Any thoughts here? Should we just go for it and commit to fixing any problems that arise quickly?

huddlej:
@tsibley I feel exactly the same way! I would prefer for this image to get more rigorous testing before we merge.

For local builds, it should be as easy as this, right?

```
# Pull the new image
docker pull nextstrain/base:branch-python-base

# Run a build with the new image
nextstrain build --image nextstrain/base:branch-python-base .
```

Maybe @lmoncla would be able to try it out since she just got ncov builds up and running. I'll ping her about this.

tsibley:
That `nextstrain` command should do it. `nextstrain` will automatically pull down the image if needed, so not necessary to run `docker pull` at all.

An important caveat is that builds run via `--aws-batch` won't use the requested custom image because of nextstrain/cli#57.

jstoja:
Hello @huddlej @tsibley,

I've just tested this PR's Docker Image with the the current `nextstrain/cli` master commit (`1.16.2-6-g8c72b6b`) and the `zika-tutorial` build (I've tried with the `ncov` one, but I'm missing credentials it seems).
It seemed to work as expected (I was able to run the `build` command, but also `view` the build in my browser). Although, it complained about narratives, but I think it is expected.

Please let me know if you have other things you need to test.

Thanks

huddlej:
Thank you for testing this out, @jstoja! Would you be willing to try the same type of test with our [mumps build](https://github.com/nextstrain/mumps)?

We think this mumps build will not work with the image defined in this PR, but it would be helpful to get confirmation about how this build is failing (assuming it does so with the example data).

jstoja:
No problem, I'll test this one too.

Small additional comment, I can see that in the `cli` repository, TravisCI is using `Ubuntu` and this Dockerfile aims to use `Debian` instead. Maybe the same should be used to avoid any future issues. Ubuntu also provides `slim` images, so something very similar could be done.

tsibley:
> Small additional comment, I can see that in the `cli` repository, TravisCI is using `Ubuntu` and this Dockerfile aims to use `Debian` instead. Maybe the same should be used to avoid any future issues. Ubuntu also provides `slim` images, so something very similar could be done.

Right, the CI tests in nextstrain/cli run in Ubuntu (e.g. the commands inside .travis.yml), but when the example build using Docker is run, it'll run inside the container, which would be Debian with this PR. I don't think this is an issue for testing, and generally we do expect that the container OS won't match the host OS. Are there particular future issues you think might crop up?

jstoja:
The only issues that I might see is the dynamic libraries being different, it might cause issues. Albeit this should be VERY rare, so probably not a huge issue.

jstoja:
> We think this mumps build will not work with the image defined in this PR, but it would be helpful to get confirmation about how this build is failing (assuming it does so with the example data).

So I've been testing this a bit more and what fails currently is an `augur filter` job that complains about group-by categories.
I assume this is related to the fact that I don't have a populated `fauna` and therefore the fasta file is empty.

I'll try to set this up and continue the testing.

tsibley:
Rebased on top of the latest `master` so I could use this image for some ad-hoc builds.

@huddlej I'd still love to pick a week to make this the default image (by merging it) and be ready to quickly turnaround hot fixes if/when issues arise. We could switch the ncov build to it earlier than that, I suppose, to try to shake out any issues before pushing it out to external users.

huddlej:
Sounds great, @tsibley! Any time Wednesday-Friday of next week should work.

tsibley:
@huddlej 

> Any time Wednesday-Friday of next week should work.

This week was unexpectedly scattered for me due to childcare, so let's plan on the coming week? My thought is to:

- [ ] Merge https://github.com/nextstrain/ncov-ingest/pull/123 later today to switch over our ncov builds

and then watch the regular builds next week on

- [ ] Monday, and
- [ ] Tuesday

If we don't see any problems, then

- [ ] Merge this PR on Tuesday/Wednesday to make it the latest default base image for everyone.
- [ ] Revert https://github.com/nextstrain/ncov-ingest/pull/123 so ncov tracks latest image again.

It would also still be good to follow up with:

https://github.com/nextstrain/docker-base/blob/71fc8bdca7880d93d173f6904ff92ccd8ceee753/Dockerfile#L96-L101

The current state of Augur has slightly different versions:

https://github.com/nextstrain/augur/blob/867c5e368e5f621528039fe5d417c85e9e66c0f0/setup.py#L59-L63

I don't think I have enough recent context here to know if bumping the Augur deps and cutting a release is all we're talking about, or if there's more to it than that. Can you weigh in?

tsibley:
(Rebased onto latest master to fix merge conflict before merge.)

huddlej:
Thank you for pushing this forward with a great plan, @tsibley! The only notable version difference in this dependencies is the cvxopt version. I've tested cvxopt with versions 1.2.X and actually use versions 1.X in [the Augur bioconda recipe](https://bioconda.github.io/recipes/augur/README.html). I can update the Augur setup.py to reflect this.

Is there a Docker-specific reason why we don't just run `pip install .[full]` when we install Augur in the image instead of installing each of these dependencies separately? I'd love to get as close as possible to one (or two?) sources of truth for dependencies, so we don't miss updating places like this. If there is a slight extra cost in Docker layer size/turnover from the full pip install, that might be worth the reduced complexity.

tsibley:
> I've tested cvxopt with versions 1.2.X and actually use versions 1.X in the Augur bioconda recipe. I can update the Augur setup.py to reflect this.

Excellent!

> Is there a Docker-specific reason why we don't just run pip install .[full] when we install Augur in the image instead of installing each of these dependencies separately? I'd love to get as close as possible to one (or two?) sources of truth for dependencies, so we don't miss updating places like this. If there is a slight extra cost in Docker layer size/turnover from the full pip install, that might be worth the reduced complexity.

With the update to Augur's `setup.py` and this PR for the new base image, I think we should definitely just install `.[full]`.

That original reason Augur's deps were explicitly denormalized into the Dockerfile here was to avoid lengthy image rebuilds. That was done because several of the deps didn't have wheels for Alpine, so had to be compiled from source. With the new image base, we should be able to rely on wheels, so installing the deps every time won't be much of a burden and worth the simplicity of having a single source of truth.

huddlej:
> That was done because several of the deps didn't have wheels for Alpine

Perfect! It's all coming full-circle in my mind. :)

tsibley:
Making the switch to `augur[full]` in https://github.com/nextstrain/docker-base/pull/30.

Dockerfile
```diff
@@ -127,41 +93,37 @@ RUN pip3 install --requirement=/nextstrain/fauna/requirements.txt
...
+
 # Install Node deps, build Auspice, and link it into the global search path.  A
 # fresh install is only ~40 seconds, so we're not worrying about caching these
 # as we did the Python deps.  Building auspice means we can run it without
 # hot-reloading, which is time-consuming and generally unnecessary in the
 # container image.  Linking is equivalent to an editable Python install and
 # used for the same reasons described above.
-RUN cd /nextstrain/auspice && npm install && npm run build && npm link
-
+RUN cd /nextstrain/auspice && npm update && npm install && npm run build && npm link
```
tsibley, line 109: @huddlej I was trying to answer "Why do we run `npm update`?" (when working on #165) and traced its introduction back to here. I assume it's unlikely, but any recollection why you added it?

huddlej: @tsibley I'm sorry I don't remember. Since I know very little about the npm ecosystem, it seems likely that I saw this pattern somewhere during work on this PR and copied it while trying to debug build issues.

tsibley: No worries!

PR #22, opened 2020-03-17
Upgrade MAFFT to the latest version, v7.450

tsibley:
Good to stay current but also curious to see if multi-threading
performance is better.


tsibley:
although [the changelog](https://mafft.cbrc.jp/alignment/software/changelog.html) doesn't make it seem promising that threading is any different.

PR #23, opened 2020-03-18
Install Snakemake with support for its reports feature

tsibley:
This allows `snakemake --report` to be used, which is nice for
generating summaries of workflow runs.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,6 +21,7 @@ RUN apk add --no-cache \
         freetype-dev \
         git \
         gmp-dev \
+        graphviz-dev \
         libpng-dev \
         linux-headers \
         nodejs \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -96,7 +97,7 @@ RUN pip3 install phylo-treetime==0.6.4
 RUN pip3 install requests==2.20.0
 RUN pip3 install rethinkdb==2.3.0.post6
 RUN pip3 install seaborn==0.9.0
-RUN pip3 install snakemake==5.8.1
+RUN pip3 install snakemake[reports]==5.8.1
```
fanninpm, line 100: Is there any particular reason for specifying an old (released 2019-11-15) version of Snakemake? As far as I can tell, the latest version of Snakemake (v7.3.0, released 2022-03-21) has support for Python 3.7.

tsibley: This isn't specific to this (outdated) PR. We've chosen so far to keep the version of Snakemake relatively static to avoid backwards compat issues that we haven't addressed yet in the [Nextstrain CLI](https://github.com/nextstrain/cli), such as the new requirement of Snakemake's `--cores` (and likely others). We should indeed update it at some point.

tsibley:
Closing this as it was a quick idea but one that wasn't needed.

PR #25, opened 2020-05-24
Update snakemake to support --set-threads argument

huddlej:
The current Docker image has Snakemake version 5.8.1, but when running builds on AWS, we need access to the `--set-threads` argument that was added in version 5.10.0. This PR updates the Docker image to this later version of Snakemake which also happens to be the latest version currently supported by augur.


tsibley:
Thanks, @huddlej! I reviewed the Snakemake changelog and no backwards incompatible changes are documented between 5.8.1 and 5.10.0, so this looks safe to me.

PR #30, opened 2021-02-05
Install Augur's "full" extra deps automatically instead of inlining them separately

tsibley:
The only notable difference at the moment is the version of cvxopt,
which is 1.2.x here, but 1.1.x in Augur's setup.py at the moment.
@huddlej notes that the Augur bioconda recipe actually uses 1.x and
Augur's setup.py should be separately updated to reflect a more lenient
version.

That original reason Augur's deps were explicitly inlined into the
Dockerfile here was to avoid lengthy image rebuilds.  That was done
because several of the deps didn't have wheels for Alpine, so had to be
compiled from source.  With the new Debian image base, we should be able
to rely on wheels, so installing the deps every time won't be much of a
burden and very much worth the simplicity of having a single source of
truth.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -91,14 +91,7 @@ RUN pip3 install --requirement=/nextstrain/fauna/requirements.txt
 # Augur is an editable install so we can overlay the augur version in the image
 # with --volume=.../augur:/nextstrain/augur and still have it globally
 # accessible and importable.
-RUN pip3 install --editable /nextstrain/augur
-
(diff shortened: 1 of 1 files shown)


tsibley:
Looks like we need the cvxopt bump in Augur's setup.py first as there aren't Python 3.7 wheels on PyPI until cvxopt 1.2.1.

victorlin (approved):

Issue #33, opened 2021-06-03
Add Pangolin tools


huddlej:
## Context
[We recently added support for reannotating Pangolin lineages as part of the ncov workflow](https://github.com/nextstrain/ncov/pull/593/files). To support these annotations, we needed to install `pangolin=2.3.8` and `pango-learn=2021.04.01` in the Snakemake Conda environment.

In a previous Nextstrain meeting, we discussed whether to include Pangolin tools in this base image or not given that these tools only work for the ncov workflow and are not needed for any other workflow that uses the "base" image. The consensus at the time was to install Pangolin tools in the base image to make maintenance easier and avoid supporting yet another image that is specific to the ncov workflow.

## Description

To support Pangolin annotations from the ncov workflow, we need to [install pangolin and pango-learn](https://cov-lineages.org/resources/pangolin/installation.html). The creators of Pangolin recommend installation with Conda, but we do not currently support Conda in the Docker image, so we need to install Pangolin and its dependencies manually based on the contents of [the most recent Conda environment file](https://github.com/cov-lineages/pangolin/blob/master/environment.yml).

## Testing

If the software is installed properly, we should be able to run an ncov workflow with `run_pangolin: true` in a `builds.yaml` configuration.

huddlej:
If we do decide we are ok with:

1. installing pathogen-specific tools in the base image
2. installing tools from Conda in the base image

Then we could follow the pattern used in [Snakemake's Dockerfile](https://github.com/snakemake/snakemake/blob/main/Dockerfile), as an example. This approach would need to keep Conda installed in the final image, so the image would be a bit bigger than it is now. On the plus side, I think this inclusion of Conda in the base image would allow users to run their Snakemake workflows with `--use-conda` (as attempted in #32).

tsibley:
Snakemake's approach for activating a default Conda env in the image using `.bashrc` will probably not work for us because we don't execute everything through bash in "interactive" mode. The setting of `PATH` will get most of the way there, but not fully-support all Conda packages as some require setting additional env vars (e.g. to shared library paths) which only happens via `conda activate …`.

tsibley:
On the other hand, Pangolin looks readily installable as a set of Python packages not via Conda, so that would seem to be simpler to me.

huddlej:
Assigning @sidneymbell as an indication that her team at CZI has volunteered to work on this in an upcoming sprint.

PR #34, opened 2021-08-12
Add epiweeks as a workflow-specific dependency

huddlej:

The ncov workflow annotates epiweeks from dates to more effectively communicate dates with epidemiologists and allow epi users to color and filter data by these annotations. This commit adds the Python `epiweeks` package as a dependency, to support epiweek annotations for ncov workflow users who rely on the Docker image.

## Related issue(s)

Required for https://github.com/nextstrain/ncov/pull/703
Related to https://github.com/nextstrain/docker-base/issues/33

## Testing

 - [x] CI

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -100,6 +100,15 @@ RUN pip3 install cvxopt==1.2.4
 RUN pip3 install matplotlib==2.2.2
 RUN pip3 install seaborn==0.9.0
 
+# Install pathogen-specific workflow dependencies. Since we only maintain a
+# single Docker image to support all pathogen workflows, some pathogen-specific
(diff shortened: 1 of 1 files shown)


tsibley (approved):
I wish this wasn't necessary, but right now it is. Merge away!

I think it'd be good to address this issue of workflow deps more generally. After my last tangle with workflow-specific deps, I [jotted down some quick notes](https://github.com/tsibley/blab-standup/blob/master/2021-01-19.txt) with ideas that had been living only in my head. Buildpacks seem to me to be a very promising approach. I'd love to chat with you about it sometime, if you're interested.

huddlej:
Thanks! I would love to chat about this more!

PR #37, opened 2022-01-11
Add Pangolin to nextstrain-base docker image

### LINKED ISSUE #33: Add Pangolin tools

jgadling:
This adds Pangolin to nextstrain-base docker image. I'm not sure specifically which version of pangolin we want to include, so I used the versions mentioned in the GitHub issue. Just let me know if I should update this!

### Related issue(s)  

Fixes #33 

### Testing
I've rebuilt the docker image and Pangolin runs successfully:

```
% ./devel/build
<output omitted>
% docker run -ti --rm=true nextstrain/base -- sh -c 'pangolin /usr/local/lib/python3.7/site-packages/pangolin/data/reference.fasta 2> /dev/null && echo output: && cat lineage_report.csv'
Found the snakefile
Do things with sam files

Usage:
  gofasta sam [flags]
  gofasta sam [command]

Available Commands:
  indels       Parse a SAM file for raw indel information
  toMultiAlign Convert a SAM file to a multiple alignment in fasta format
  toPairAlign  convert a SAM file to pairwise alignments in fasta format
  variants     Find mutations relative to a reference from an alignment in sam format

Flags:
  -h, --help               help for sam
  -r, --reference string   Reference fasta file used to generate the sam file
  -s, --samfile string     Samfile to read. If none is specified, will read from stdin (default "stdin")
  -t, --threads int        Number of threads to use (default 1)

Use "gofasta sam [command] --help" for more information about a command.
2.24-r1122
5.10.0
The query file is:/usr/local/lib/python3.7/site-packages/pangolin/data/reference.fasta
Looking in /usr/local/lib/python3.7/site-packages/pangoLEARN/data for data files...

Data files found
Trained model:  /usr/local/lib/python3.7/site-packages/pangoLEARN/data/decisionTree_v1.joblib
Header file:    /usr/local/lib/python3.7/site-packages/pangoLEARN/data/decisionTreeHeaders_v1.joblib
Lineages csv:   /usr/local/lib/python3.7/site-packages/pangoLEARN/data/lineages.metadata.csv
loading model 01/11/2022, 02:07:00
processing block of 1 sequences 01/11/2022, 02:07:01
complete 01/11/2022, 02:07:02
Output file written to: /nextstrain/build/lineage_report.csv
output:
taxon,lineage,probability,pangoLEARN_version,status,note
outgroup_A,A,1.0,2021-04-01,passed_qc,
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -185,7 +185,6 @@ COPY --from=builder /usr/local/lib/python3.7/site-packages/ /usr/local/lib/pytho
 # troublesome or excessive.
 #   -trs, 15 June 2018
 COPY --from=builder \
-    /usr/local/bin/*.py \
     /usr/local/bin/augur \
     /usr/local/bin/aws \
     /usr/local/bin/envdir \
--- a/Dockerfile
+++ b/Dockerfile
@@ -127,6 +127,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         bzip2 \
         ca-certificates \
         curl \
+        git \
         gzip \
         nodejs \
         perl \
@@ -150,6 +151,15 @@ COPY --from=builder \
 
 COPY --from=builder /build/vcftools/built/bin/    /usr/local/bin/
 COPY --from=builder /build/vcftools/built/share/  /usr/local/share/
+
+# Add Pangolin and PangoLEARN + deps
+RUN curl -fsSL https://github.com/virus-evolution/gofasta/releases/download/v0.0.6/gofasta-linux-amd64 \
+        -o /usr/local/bin/gofasta \
+  && chmod a+rx /usr/local/bin/gofasta
+RUN cd /usr/local/bin && curl -fsSL https://github.com/lh3/minimap2/releases/download/v2.24/minimap2-2.24_x64-linux.tar.bz2 \
+  | tar xjvpf - --strip-components=1 minimap2-2.24_x64-linux/minimap2
+RUN pip install git+https://github.com/cov-lineages/pangolin.git@v2.3.8
+RUN pip install git+https://github.com/cov-lineages/pangoLEARN.git@2021-04-01
 
 # Add Nextalign
 RUN curl -fsSL https://github.com/nextstrain/nextclade/releases/latest/download/nextalign-Linux-x86_64 \
--- a/Dockerfile
+++ b/Dockerfile
@@ -148,7 +148,6 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         xz-utils \
         zip unzip
 
-
 # Configure the prompt for interactive usage
 COPY prompt.sh /etc/profile.d/
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -132,45 +132,6 @@ RUN cd /nextstrain/auspice && npm update && npm install && npm run build && npm 
 
 # ———————————————————————————————————————————————————————————————————— #
 
-# UShER is a pangolin 3+ dependency.
-FROM python:3.7-slim-buster as usher
-ENV DEBIAN_FRONTEND=noninteractive
-RUN apt-get update && apt-get install -yq --no-install-recommends \
-    automake \
-    build-essential \
-    cmake \
-    curl \
-    ca-certificates \
-    libboost-date-time-dev \
-    libboost-filesystem-dev \
-    libboost-iostreams-dev \
-    libboost-program-options-dev \
-    libmpich-dev \
-    libprotoc-dev \
-    libprotoc-dev protobuf-compiler \
-    libtbb-dev \
-    libtool \
-    mafft \
-    mpich \
-    nasm
-
-WORKDIR /usr/local/lib/
-RUN curl -fsSL https://github.com/intel/isa-l/archive/refs/tags/v2.30.0.tar.gz \
-  | tar xzvpf -
-WORKDIR /usr/local/lib/isa-l-2.30.0
-RUN ./autogen.sh && ./configure && make -j 2 && make install
-WORKDIR /usr/local/lib/
-RUN curl  -fsSL https://github.com/oneapi-src/oneTBB/archive/2019_U9.tar.gz \
-  | tar xzvpf -
-RUN curl -fsSL https://github.com/yatisht/usher/archive/refs/tags/v0.5.2.tar.gz \
-  | tar xzvpf -
-WORKDIR /usr/local/lib/usher-0.5.2
-## Checkout latest release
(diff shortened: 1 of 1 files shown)


tsibley (changes requested):
Thanks for this PR, @jgadling!

This Dockerfile uses a [two-stage build](https://docs.docker.com/develop/develop-images/multistage-build/) with most of the work happening in the first stage `builder` image. The second stage then copies artifacts from the builder image over into the final image, keeping the final image free of things only needed for building (e.g. git, a C toolchain, etc).

The code added by this PR is in the second-stage final image but should instead be in the first-stage `builder` image. Could you rework the PR to move it? You'll also then not need the added `git` dep in the second stage.

Dockerfile
```diff
@@ -151,6 +152,15 @@ COPY --from=builder \
...
 COPY --from=builder /build/vcftools/built/share/  /usr/local/share/
 
+# Add Pangolin and PangoLEARN + deps
+RUN curl -fsSL https://github.com/virus-evolution/gofasta/releases/download/v0.0.6/gofasta-linux-amd64 \
+        -o /usr/local/bin/gofasta \
+  && chmod a+rx /usr/local/bin/gofasta
+RUN cd /usr/local/bin && curl -fsSL https://github.com/lh3/minimap2/releases/download/v2.24/minimap2-2.24_x64-linux.tar.bz2 \
+  | tar xjvpf - --strip-components=1 minimap2-2.24_x64-linux/minimap2
+RUN pip install git+https://github.com/cov-lineages/pangolin.git@v2.3.8
+RUN pip install git+https://github.com/cov-lineages/pangoLEARN.git@2021-04-01
```
huddlej, line 162: The pango versions mentioned in the original issue that this PR addresses are pretty old now. I think we'll want the latest versions of `pangolin` ([3.1.17](https://github.com/cov-lineages/pangolin/releases/tag/v3.1.17)) and pangoLEARN ([2021-12-06](https://github.com/cov-lineages/pangoLEARN/releases/tag/2021-12-06)).

It looks like [pangolin version 3.0+ requires UShER](https://cov-lineages.org/resources/pangolin/installation.html), too. UShER provides [an installation script for Ubuntu](https://github.com/yatisht/usher/blob/master/install/installUbuntu.sh) that we could use as a starting point.

huddlej (commented):
Thank you for putting this together, @jgadling and @sidneymbell! My only additional comment to Tom's is about trying this out with the latest pango versions. It seems we'll need to install UShER for the latest pangolin, but let us know if this is more work than you can commit to at the moment. We can pick up from this PR.

Dockerfile
jgadling: Ok - I've managed to get Pangolin 3.x with UShER functionality working in the image build, but if I'm reading the Pango docs correctly, UShER is an *optional* dependency for pangolin ([docs](https://cov-lineages.org/resources/pangolin/usage.html)) and I don't see anything in nexstrain that enables that functionality, so I'm omitting it.

sidneymbell:
Hey there @tsibley and @huddlej -- is this branch now good to merge? :)

Dockerfile
```diff
@@ -172,10 +185,16 @@ COPY --from=builder /usr/local/lib/python3.7/site-packages/ /usr/local/lib/pytho
 # troublesome or excessive.
 #   -trs, 15 June 2018
 COPY --from=builder \
+    /usr/local/bin/*.py \
```
huddlej, line 188: I tried running pango annotations from the ncov workflow with this line commented out and everything worked as expected.

Are there specific Python scripts we should include here? Even if there are a lot, an explicit list would help with future maintenance.

huddlej (commented):
Thank you for the updates, @jgadling! You were absolutely right about not needing UShER for our ncov use case; it's nice to avoid that extra complexity in this file.

I tested this Dockerfile like so:

```
# Pull and build.
./devel/pull
./devel/build

# Run pango annotations in ncov workflow with Nextstrain CI build.
cd ../ncov
nextstrain build --cpus 2 --image nextstrain/base:latest . -np --forceall --configfile nextstrain_profiles/nextstrain-ci/builds.yaml --config run_pangolin=True
```

This all worked mostly as expected, although I got a couple of scikit-learn warnings that I hadn't noticed in our ncov conda environment before. These warnings seem to stem from using a later version of scikit-learn in our Docker image that was used to prepare the pango model files:

```
/usr/local/lib/python3.7/site-packages/sklearn/base.py:338: UserWarning: Trying to unpickle estimator DecisionTreeClassifier from version 0.24.2 when using version 1.0.2. This might lead to breaking code or invalid results. Use at your own risk. For more info please refer to:
https://scikit-learn.org/stable/modules/model_persistence.html#security-maintainability-limitations
  UserWarning,
/usr/local/lib/python3.7/site-packages/sklearn/base.py:444: UserWarning: X has feature names, but DecisionTreeClassifier was fitted without feature names
  f"X has feature names, but {self.__class__.__name__} was fitted without"
```

I visually confirmed that the pango annotations looked correct, though, and there were no errors. I haven't dug into why we have scikit-learn installed, but these warnings don't seem associated with a real problem.

I made a note below about this, but I didn't see a specific need to copy any additional Python scripts from `/usr/local/bin` in the second stage of the image build. Were there specific scripts you found were required in your own testing? If not, we can drop that line and merge the rest of the changes as they are.

Dockerfile
jgadling: This could have been left over from the attempt to install usher, you're right that I don't think it's necessary. I've removed it.

huddlej (approved):
Looks good to me, @jgadling! Thank you again for your work on this issue!

PR #44, opened 2022-05-23
Install Augur from the latest stable version tag instead of the "release" branch

tsibley:
The "release" branch was recently removed.¹  Instead, use the Git tag
for the latest stable version released to PyPI.

There's a larger discussion to be had about if it makes any sense any
more for Augur to be installed from Git instead of PyPI (and
correspondingly, Auspice from NPM), but that's for another time.

¹ https://github.com/nextstrain/augur/pull/913

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,6 +18,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         ca-certificates \
         curl \
         git \
+        jq \
         libgmp-dev \
         libpng-dev \
         nodejs \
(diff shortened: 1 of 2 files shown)


huddlej (commented):
Thank you for catching this, @tsibley! I'm sorry we missed it in that PR you linked to.

Do you prefer the PyPI API for version info over GitHub's API to avoid grabbing a potential non-release tag from GitHub? In other words, we no longer know what the "stable" version is for Augur based on the GitHub tags alone, but we need to query PyPI as a kind of external validation?

It is a little confusing to have two different sources of truth (one for version info and one for the software itself), but I understand that conversation being outside the scope of this PR.

victorlin (approved):

tsibley:
@huddlej 

Yeah, I chose to query PyPI because it both handles the pre-release issue for us and is upstream of Pip and Conda (by way of Bioconda). PyPI in many senses de-facto defines the latest stable release.

To put it another way, if there's a stable release on PyPI, then it's what's installed by default by a number of mechanisms. But if there's a stable release on GitHub Releases, there's nothing that installs it by default. GitHub Releases doesn't really make sense as a _release distribution_ mechanism for Augur, though it does make sense as a _release announcement/information_ mechanism.

So Docker being downstream of PyPI just like Pip and Conda makes sense to me. Especially since, as I hinted, I think we may want to stop installing Augur (and Auspice) into the image from Git anyway and switch to installing from PyPI (and NPM).

PR #46, opened 2022-06-02
Use GitHub Actions for build/pull/tag/push

### LINKED ISSUE #35: Build multi-arch image with amd64 + arm64 for M1 Macs

Reference material:

victorlin:

This replaces the existing scripts with GitHub Actions provided by Docker.

Also has the benefit of using buildx which supports building multi-arch images (to be done separately in #48): https://docs.docker.com/desktop/multi-arch/

Sequence of steps inspired by https://github.com/docker/build-push-action/blob/master/docs/advanced/isolated-builders.md.

### Related issue(s)

- Preparing for #35

### Testing

See CI runs. However, this doesn't test the execution path for `push` trigger on the default branch `master`.


.github/workflows/ci.yml
```diff
@@ -5,24 +5,98 @@ jobs:
...
-        ./devel/tag $tag
-        ./devel/push $tag
+
+    # The nextstrain/base Dockerfile is a multi-stage with both a "builder" target
+    # and a main target.  To enable proper caching via --cache-from we need both
+    # these images available to pull layers from.  This means pulling both in at
+    # the start and pushing both up at the end.
+
+    # Calling `docker run nextstrain/base` will still only pull down the small base
+    # image rather than pulling down the larger nextstrain/base-builder image.
```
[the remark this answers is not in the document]
tsibley: I believe one of the features of buildx is better support for multi-stage builds, such that you can do just `build` operation for a multi-stage build but still push up tagged images for each stage (instead of having to explicitly `build` each stage yourself just so you can push each later).

I'm not sure how building both stages in parallel would work since the second stage relies on the first?

.github/workflows/ci.yml
```diff
@@ -5,24 +5,98 @@ jobs:
...
+    - name: Show base-image-builder name
+      run: echo ${{ steps.base-image-builder.outputs.name }}
+
+    # Set CACHE_DATE in your environment to force layers after our custom cache
+    # point to be re-built. See the ARG CACHE_DATE line in the Dockerfile for more
+    # information.
+    - name: Set CACHE_DATE and GIT_REVISION env vars
+      run: |
+        echo "CACHE_DATE=$(date --utc +%Y%m%dT%H%M%SZ)" >> $GITHUB_ENV
+        echo "GIT_REVISION=$(git describe --tags --abbrev=40 --always --dirty || true)" >> $GITHUB_ENV
```
tsibley, line 27: I like to do `| tee -a` instead of `>>` so that the values are visible in the workflow logs.

.github/workflows/ci.yml
```diff
@@ -5,24 +5,98 @@ jobs:
...
+      name: Set tag name (non-default branch)
+      run: echo "TAG_NAME=branch-${GITHUB_REF_NAME//[^A-Za-z0-9._-]/-}" >> $GITHUB_ENV
+
+    - if: ${{ fromJSON(env.TAG_AND_PUSH) }}
+      name: Set tags to push
+      run: |
+        maybe_base_image_latest_tag=${{ fromJSON(env.ON_DEFAULT_BRANCH) && ',$BASE_IMAGE_NAME:latest' || '' }}
+        echo "BASE_IMAGE_TAGS=$BASE_IMAGE_NAME:$TAG_NAME$maybe_base_image_latest_tag" >> $GITHUB_ENV
+        maybe_builder_image_latest_tag=${{ fromJSON(env.ON_DEFAULT_BRANCH) && ',$BUILDER_IMAGE_NAME:latest' || '' }}
+        echo "BUILDER_IMAGE_TAGS=$BUILDER_IMAGE_NAME:$TAG_NAME$maybe_builder_image_latest_tag" >> $GITHUB_ENV
```
tsibley, line 56: This is all pretty dense and hard for me to follow given the mix of interpolation at different levels, JSON serde, and conditionals. The motivation for defining and using all these is because the `build-push-action`s needed to be parameterized instead of themselves being in a conditional like with the modular `devel/tag` + `devel/push` commands?

.github/workflows/ci.yml
```diff
@@ -5,24 +5,98 @@ jobs:
...
+        target: builder
+        builder: ${{ steps.builder-image-builder.outputs.name }}
+        platforms: linux/amd64
+        context: .
+        pull: true
+        push: ${{ fromJSON(env.TAG_AND_PUSH) }}
+        tags: ${{ fromJSON(env.TAG_AND_PUSH) && env.BUILDER_IMAGE_TAGS || null }}
+        cache-from: |
+          type=registry,ref=${{ env.BUILDER_IMAGE_NAME }}
+          type=registry,ref=${{ env.BASE_IMAGE_NAME }}
```
tsibley, line 83: The lack of `cache-to` here might mean that the subsequent step can't use the just-built first stage when `TAG_AND_PUSH` is false (and would have to re-download it with `TAG_AND_PUSH` is true).

.github/workflows/ci.yml
```diff
@@ -5,24 +5,98 @@ jobs:
...
+      with:
+        builder: ${{ steps.base-image-builder.outputs.name }}
+        platforms: linux/amd64
+        context: .
+        pull: true
+        push: ${{ fromJSON(env.TAG_AND_PUSH) }}
+        tags: ${{ fromJSON(env.TAG_AND_PUSH) && env.BASE_IMAGE_TAGS || null }}
+        cache-from: |
+          type=registry,ref=${{ env.BUILDER_IMAGE_NAME }}
+          type=registry,ref=${{ env.BASE_IMAGE_NAME }}
```
tsibley, line 99: Similarly, the lack of non-registry sources in this `cache-from` here might mean that this step can't use the just-built first stage from the previous step when `TAG_AND_PUSH` is false (and would have to re-download it with `TAG_AND_PUSH` is true).

.github/workflows/ci.yml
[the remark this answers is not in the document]
tsibley: Ah, this better support appears to be ([1](https://github.com/docker/buildx/blob/master/docs/reference/buildx_build.md#-export-build-cache-to-an-external-cache-destination---cache-to), [2](https://github.com/moby/buildkit#export-cache)) in the form of exporting cached layers via `type=registry` with `mode=max` (meaning "all intermediate layers"), but you have to use a separate image tag/name (unlike `type=inline`, but that only supports `mode=min` which means "final layers").

tsibley (changes requested):
+1 for using buildx here, but as-is I'm not sure this seems like an improvement to me over the modular shell scripts. In particular, I think the GitHub Actions flow is much harder to follow, and it seems unfortunate that there's no longer a shared toolchain for building the images locally.

I wonder about instead ditching `build-push-action` here but keeping `setup-buildx` and then keeping the modular programs in `devel/` but extending them to use `docker buildx build …` and pass `--builder …` and `--platforms …`. (Alternatively, I think `docker buildx build` could even be made optional by setting `install: true` for `setup-buildx`, but not sure that's a good idea or not.)

.github/workflows/ci.yml
victorlin: Yeah, I couldn't figure out a better way since GitHub Actions provides limited support for conditional "code paths" here. It would be either some variant of this (with comments to explain) or several copies of `build-push-action` that are mostly the same except for `push`, `tags`, and step condition.

victorlin:
> @tsibley: I wonder about instead ditching `build-push-action` here but keeping `setup-buildx` and then keeping the modular programs in `devel/` but extending them to use `docker buildx build …`

That sounds reasonable, it addresses the motivation for this with much less of a change.

PR #48, opened 2022-06-02
Add `arm64` to image builds

### LINKED ISSUE #35: Build multi-arch image with amd64 + arm64 for M1 Macs

victorlin:
### Related issue(s)

- Fixes #35

### Testing

See checks.

### TODO

- [ ] Fix compiling issues in `arm64` build (see checks)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,8 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
+    - uses: docker/setup-qemu-action@v2
+
     - uses: docker/setup-buildx-action@v1
       id: builder-image-builder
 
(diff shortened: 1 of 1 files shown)


tsibley:
:eyes:

```
RUN make -f Makefile.AVX.PTHREADS.gcc   # AVX should be widely-supported enough
…
gcc: error: unrecognized command line option ‘-msse3’
```

…widely-supported enough _by the x86 family of chips_!  :upside_down_face:

tsibley:
Guess we'll need to not compile FastTree with x86 instruction set features on non-x86 chips.

victorlin:
> @tsibley: Guess we'll need to not compile FastTree with x86 instruction set features on non-x86 chips.

I think the culprit is RAxML, but yeah

https://github.com/nextstrain/docker-base/blob/163cba5590d615d089c0c881077eb5033a9929e0/Dockerfile#L35-L41

tsibley:
Oh oops I misread, totally RAxML not FastTree.

PR #49, opened 2022-06-14
Use buildx create+build

### LINKED ISSUE #70: CI should fail when pushing image with identical digest
*Identified by @tsibley in https://github.com/nextstrain/docker-base/pull/49#issuecomment-1175410538.*

In #49, some changes seemingly worked with a successful push, but the lack of `docker buildx --output` meant that the pulled image was being pushed, not the image produced by `docker buildx`. CI should fail in this scenario by comparing the digest of the remote `latest` image with the to-be-pushed digest of the newly built image.

To get the digest for the remote `nextstrain/base:latest`:

```sh
docker manifest inspect nextstrain/base:latest -v | jq -r '.Descriptor.digest'
```

I had troubles getting the digest for the newly built image. Here's what I've tried:

```sh
# This is a digest, but not what's actually pushed by `docker push`. Not sure why.
docker images --no-trunc --quiet nextstrain/base:$TAG

# This is a digest which is different from the above, so maybe the right one?
# However, this doesn't work on builds from `docker build`, but works on builds from `docker buildx`.
# See failure with `docker build`: https://github.com/nextstrain/docker-base/runs/7242000364?check_suite_focus=true
docker inspect --format='{{index .RepoDigests 0}}' nextstrain/base:$TAG
```

victorlin:

This makes it easier to support multi-platform builds in the future, and
explicitly sets the currently used platform (linux/amd64).

Motivated by https://github.com/nextstrain/docker-base/pull/46#pullrequestreview-1005012188, but using `docker buildx create` in the script instead of `setup-buildx-action`.

### Related issue(s)

- Supersedes/closes #46
    - closes #48
- Preparing for #35
- Closes #70

### Testing

See checks.

### DIFF
--- a/devel/build
+++ b/devel/build
@@ -19,19 +19,37 @@ export GIT_REVISION=$(git describe --tags --abbrev=40 --always --dirty || true)
 # Calling `docker run nextstrain/base` will still only pull down the small base
 # image rather than pulling down the larger nextstrain/base-builder image.
 
-docker build \
+platform=linux/amd64
+
+# `buildx create` is necessary to use a driver that supports multi-platform
+# images.
+builder=nextstrain-builder
+
+if ! docker buildx inspect "$builder" &>/dev/null; then
+    docker buildx create --name "$builder" --driver docker-container
+fi
+
+docker buildx build \
+    --builder "$builder" \
+    --platform $platform \
     --build-arg CACHE_DATE \
     --build-arg GIT_REVISION \
     --cache-from nextstrain/base-builder \
     --cache-from nextstrain/base \
+    --cache-to type=inline \
     --tag nextstrain/base-builder \
+    --load \
     --target builder \
     .
 
-docker build \
+docker buildx build \
+    --builder "$builder" \
+    --platform $platform \
     --build-arg CACHE_DATE \
     --build-arg GIT_REVISION \
     --cache-from nextstrain/base-builder \
     --cache-from nextstrain/base \
+    --cache-to type=inline \
     --tag nextstrain/base \
+    --load \
     .
--- a/devel/tag
+++ b/devel/tag
@@ -2,6 +2,7 @@
 #
 # Assign the given additional tags to the nextstrain/base:latest and
 # nextstrain/base-builder:latest images.
+# Errors if a tagged image has already been pushed.
 #
 set -euo pipefail
 
@@ -10,7 +11,15 @@ if [[ $# -eq 0 ]]; then
     exit 1
 fi
 
+BASE_IMAGE="nextstrain/base"
+BASE_BUILDER_IMAGE="nextstrain/base-builder"
+
 for tag in "$@"; do
-    docker tag nextstrain/base-builder:{latest,$tag}
-    docker tag nextstrain/base:{latest,$tag}
+    docker tag $BASE_BUILDER_IMAGE:{latest,$tag}
+    docker tag $BASE_IMAGE:{latest,$tag}
 done
+
+if [[ $(docker image inspect --format "{{.RepoDigests}}" $BASE_IMAGE:$tag) != '[]' || $(docker image inspect --format "{{.RepoDigests}}" $BASE_BUILDER_IMAGE:$tag) != '[]' ]]; then
+    echo "At least one of the tagged images has already been pushed. This can happen if the newly built image is not available in the local registry." >&2
+    exit 1
+fi


devel/build
```diff
@@ -19,7 +19,14 @@ export GIT_REVISION=$(git describe --tags --abbrev=40 --always --dirty || true)
 # Calling `docker run nextstrain/base` will still only pull down the small base
 # image rather than pulling down the larger nextstrain/base-builder image.
 
-docker build \
+platform=linux/amd64
+
+# `buildx create` is necessary to use a driver that supports multi-platform
+# images.
+docker buildx create --use
```
tsibley, line 26: Ah, this will create a new builder every time it's run, which is ok for the fresh envs of a CI, but not for local dev.

tsibley (changes requested):

tsibley:
Switching to buildx should also give us better caching options to reduce build times, but that can be a separate PR since we also want buildx for multi-platform builds.

devel/build
victorlin: @tsibley how about b96b606625360eed3a18abde2238c3aff58f8f7b?

devel/build
```diff
@@ -19,7 +19,14 @@ export GIT_REVISION=$(git describe --tags --abbrev=40 --always --dirty || true)
 # Calling `docker run nextstrain/base` will still only pull down the small base
 # image rather than pulling down the larger nextstrain/base-builder image.
 
-docker build \
+platform=linux/amd64
+
+# `buildx create` is necessary to use a driver that supports multi-platform
+# images.
+builder=$(docker buildx create --use)
```
tsibley, line 26: Since we're capturing the builder name here, we don't need `--use`. And it's nice not to make a temporary global state change on the user, as that could have other impacts.

```suggestion
builder=$(docker buildx create)
```

devel/build
```diff
@@ -19,7 +19,14 @@ export GIT_REVISION=$(git describe --tags --abbrev=40 --always --dirty || true)
...
+
+docker buildx build \
+    --platform $platform \
```
tsibley, line 29: …and correspondingly, we should pass the name to `--builder` here.

```suggestion
docker buildx build \
    --builder "$builder" \
    --platform $platform \
```

devel/build
```diff
@@ -28,10 +35,13 @@ docker build \
     --target builder \
     .
 
-docker build \
+docker buildx build \
+    --platform $platform \
```
tsibley, line 39: …and here as well of course.

```suggestion
docker buildx build \
    --builder "$builder" \
    --platform $platform \
```

devel/build
```diff
@@ -19,7 +19,14 @@ export GIT_REVISION=$(git describe --tags --abbrev=40 --always --dirty || true)
 # Calling `docker run nextstrain/base` will still only pull down the small base
 # image rather than pulling down the larger nextstrain/base-builder image.
 
-docker build \
+platform=linux/amd64
+
+# `buildx create` is necessary to use a driver that supports multi-platform
+# images.
```
tsibley, line 25: At first glance I didn't understand why we need to always `create`, but this comment helped it make sense, so that's nice. :-) (Assuming I understand correctly that the default `docker` builder is insufficient compared to the `docker-container` builder that `create` gives us by default.)

tsibley (changes requested):
A couple minor things, but otherwise looks good.

tsibley:
I realized that this repo doesn't run any integration tests on the new images produced by CI, so it's possible that they can build fine but not actually function in real usage. I guess we should fix that in the future! But for now, and mostly for kicks, I ran the zika-tutorial with this buildx-produced image as a very cursory check:

```
nextstrain build --image nextstrain/base:branch-victorlin-use-buildx-keep-scripts zika-tutorial -F
```

It ran from start to finish just fine.

devel/build
victorlin: Done: e70bf410aa936a83a5330a89af4570ccb807cf77

devel/build
victorlin: Done: e70bf410aa936a83a5330a89af4570ccb807cf77

devel/build
victorlin: That's also my understanding, though I couldn't find explicit documentation that states:

1. `docker buildx build` uses `docker` driver by default
2. `docker buildx create` uses non-`docker` driver by default

[The `--driver` section of `docker buildx create` docs](https://docs.docker.com/engine/reference/commandline/buildx_create/#driver) is probably the most relevant.

tsibley (changes requested):
I think the `docker buildx build` invocations will need to specify `--output` ([see docs](https://docs.docker.com/engine/reference/commandline/buildx_build/#output)) so the images are available to the local Docker client/daemon in `docker images`.

tsibley:
In fact, I bet that currently this CI looks like it works but actually tags/pushes an image that's identical to what it pulls initially, since the build step doesn't load the resulting images into the local `docker images` store/repository/cache.

tsibley:
Appears so, as the [digest of the `nextstrain/base:latest` image that's pulled](https://github.com/nextstrain/docker-base/runs/7119232757?check_suite_focus=true#step:4:329) is the same as the [digest of the `nextstrain/base:branch-victorlin-use-buildx-keep-scripts` image that's pushed](https://github.com/nextstrain/docker-base/runs/7119232757?check_suite_focus=true#step:13:210).

victorlin:
Wow, thanks for catching that! I've added a commit that should catch these situations in CI. Waiting for checks to pass/fail as expected, then I'll look into adding `--output`.

devel/build
victorlin: Done:

https://github.com/nextstrain/docker-base/blob/40339783f16d3c76901bb13011467f40aaed6f34/devel/build#L26

devel/build
tsibley: I pushed a fixup commit that's a variant of that to remove the builder in more cases: 06f7c593917280f7278cdf950413b0c3701a5a15

tsibley (changes requested):
Running this locally a few times, I noticed that no caching happens. I think this is because the `--cache-from` statements in the context of `buildx` only look at the registry and not the local image cache (e.g. what `docker image ls` shows a view of). In CI this won't matter as much, as we don't expect the local image cache to be populated, but it makes the local dev loop quite a bit slower. I don't think we can make that better with BuildKit unless we cache to/from a local dir as well as the registry.

I also added a `--cache-to type=inline` option to each invocation in the fixup commit 0cbd208. This, in my understanding, preserves the implicit behaviour of non-BuildKit `docker build` and is what will make `--cache-from` the registry work. Note that we build with BuildKit, output the image into the local Docker image cache, and then push to a registry from that image cache. It's seemingly more typical to push to the registry directly from BuildKit.

tsibley:
There's a lot of subtly with buildx that we're learning as we go here! I'd put off switching to it for so long because I was wary of this, even though it's advertised as a "drop in replacement". Good to be finally getting the ducks in a row though since it will ultimately make multi-platform and better caching possible!

victorlin:
> Note that we build with BuildKit, output the image into the local Docker image cache, and then push to a registry from that image cache. It's seemingly more typical to push to the registry directly from BuildKit.

I've also noticed this and seems like it will get more complex to have with multi-platform builds locally (no `--push`). Addressing #68 should simplify things here, but that can be another PR.

devel/build
```diff
@@ -19,19 +19,33 @@ export GIT_REVISION=$(git describe --tags --abbrev=40 --always --dirty || true)
...
+trap "docker buildx rm $builder" EXIT
+
+docker buildx build \
+    --builder "$builder" \
+    --platform $platform \
     --build-arg CACHE_DATE \
     --build-arg GIT_REVISION \
     --cache-from nextstrain/base-builder \
     --cache-from nextstrain/base \
+    --cache-to type=inline \
```
victorlin, line 37: @tsibley Looking at runs from [before](https://github.com/nextstrain/docker-base/commit/9a0f6ae647e422403e3d635518a551999905b012) and [after](https://github.com/nextstrain/docker-base/commit/947a5683e61a128d5a7c1227d75f716da955debb) this was added, I can't make out the benefit of adding `--cache-to type=inline`. The only thing I notice is [the addition of 62.9s](https://github.com/nextstrain/docker-base/runs/7307286281?check_suite_focus=true#step:5:3608):

```
#48 preparing layers for inline cache
#48 DONE 62.9s
```

tsibley: Ah, this is what I was trying to get at in https://github.com/nextstrain/docker-base/pull/49#pullrequestreview-1034754833, but I probably could have been more clear.

I don't think the `--cache-from` will work as we expect without the `--cache-to` I added, as otherwise the caching info won't have been added to the image pushed to the registry and thus can't be imported from the registry later. I haven't tested this, but it's based on my general understanding of how BuildKit works and [this bit of doc about exporting the cache](https://github.com/moby/buildkit#export-cache).

Furthermore, I expect that we won't ever see a local benefit from `--cache-to type=inline` since `--cache-from` always looks at the registry and never the local images (even though we do `--load` a local image with inlined cache). But again, haven't tested this directly.

victorlin: I was confused about the intended effect here, but understand now that `--cache-to` only affects subsequent pulls of the pushed image.

I did some testing:

- Used 62516fef123e95177f2886da3fcc84d225c051d4 to pull from the image built by this PR's source branch.
- Created b330c68f34d2df26e10965f2c28d02825a56fd27 to remove the `--cache-to` and used c1510d47dbea4b5744f42acf5b1917391d2f0f34 to pull from that built image.

At first, I was still confused because both [the run on](https://github.com/nextstrain/docker-base/runs/7326750191?check_suite_focus=true#step:5:53) 62516fef123e95177f2886da3fcc84d225c051d4 and [the run on](https://github.com/nextstrain/docker-base/runs/7327333913?check_suite_focus=true#step:5:48) c1510d47dbea4b5744f42acf5b1917391d2f0f34 show the message `importing cache manifest from ...`, but I think this is due to having `--cache-from`, not necessarily meaning there is any cache available.

But looking closer, it does look like there is a lack of caching when building on an image pushed without `--cache-to`: just go to the logs and search for `RUN pip3 install nextstrain-cli==2.0.0.post1`

victorlin, line 37: > @tsibley: Running this locally a few times, I noticed that no caching happens. I think this is because the `--cache-from` statements in the context of `buildx` only look at the registry and not the local image cache (e.g. what `docker image ls` shows a view of). In CI this won't matter as much, as we don't expect the local image cache to be populated, but it makes the local dev loop quite a bit slower. I don't think we can make that better with BuildKit unless we cache to/from a local dir as well as the registry.

How did you notice this? I just ran this twice locally, and noticed the second time was near-instant with everything marked as `CACHED`:

```sh
export GIT_REVISION=$(git describe --tags --abbrev=40 --always --dirty || true)
export CACHE_DATE=$(date --utc +%Y%m%dT%H%M%SZ)
platform=linux/amd64
builder=$(docker buildx create --driver docker-container)
trap "docker buildx rm $builder" EXIT
time docker buildx build \
    --builder "$builder" \
    --platform $platform \
    --build-arg CACHE_DATE \
    --build-arg GIT_REVISION \
    --cache-from nextstrain/base-builder \
    --cache-from nextstrain/base \
    --cache-to type=inline \
    --tag nextstrain/base-builder \
    --target builder \
    .
```

tsibley: Thanks for confirming!

tsibley: Hadn't thought of it before, but your testing setup makes me think we should be adding `--cache-from` for the `branch-*` tag when applicable (in addition to the existing `--cache-from` the `latest` tag).

tsibley: Hmm, I noticed it when running `./devel/build` in succession… though maybe not _immediate_ succession since it takes a few minutes and I did some other review work while it built. It's possible I perturbed something that invalidated the cache... it started from the top again IIRC. Trying again now…

tsibley: In case it's relevant while I wait, versions of things.

<details>

```
$ docker version
Client:
 Version:           20.10.7
 API version:       1.41
 Go version:        go1.13.8
 Git commit:        20.10.7-0ubuntu5~18.04.3
 Built:             Mon Nov  1 01:04:14 2021
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true

Server:
 Engine:
  Version:          20.10.7
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.8
  Git commit:       20.10.7-0ubuntu5~18.04.3
  Built:            Fri Oct 22 00:57:37 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.5.5-0ubuntu3~18.04.2
  GitCommit:        
 runc:
  Version:          1.0.1-0ubuntu2~18.04.1
  GitCommit:        
 docker-init:
  Version:          0.19.0
  GitCommit:        

$ docker buildx version
github.com/docker/buildx v0.7.1 05846896d149da05f3d6fd1e7770da187b52a247
```

</details>

tsibley:
Two subsequent runs of `./devel/build` with no intervening changes both started from the top for me (i.e. with `apt-get update`).

In your test, did the same underlying builder get reused? (perhaps because the exit trap never fired if the same shell was used?)

I'm testing out sharing the builder now…

devel/build
tsibley: Two subsequent runs of `./devel/build` with no intervening changes both started from the top for me (i.e. with `apt-get update`).

In your test, did the same underlying builder get reused? (perhaps because the exit trap never fired if the same shell was used?)

I'm testing out sharing the builder now…

tsibley: Yep, that did it. The build cache is specific to the builder, which is unsurprising once you realize it.

See 6ae4036 for a fixup to re-use the builder between `./devel/build` invocations. This won't matter in CI, but helps a lot for local builds.

victorlin: Oh! Unsurprising but took two to catch 😅

victorlin: Added that idea to #57

tsibley (approved):
Looks all good to me now! Don't forget to drop the tip commit… ;-)

PR #50, opened 2022-06-14
Support building multi-platform images and add `linux/arm64`

### LINKED ISSUE #35: Build multi-arch image with amd64 + arm64 for M1 Macs

victorlin:

Currently, the Docker images are built for `linux/amd64` only. These changes build the Docker images for both `linux/amd64` and `linux/arm64`.

This will substantially improve the performance of the Docker runtime for Apple silicon users, as it no longer needs to use slow Intel architecture emulation. A major drawback is the increased build time since building the `linux/arm64` image on a `linux/amd64` GitHub Actions runner requires emulation. However, IMO the benefit to users outweighs the increased build-time on our end, which can be improved in the future.

Cross-compilation (#98) is a major build-time improvement on top of this PR to be considered separately.

### Changes made outside of PR

- [x] Added `docker/setup-qemu-action@*` to allowed third party actions under @nextstrain

### Related issue(s)

- Resolves #35
- Supercedes/closes #48

### Testing

- [x] Checks pass
- [x] Multi-arch images for the PR branch are available on Docker Hub for [nextstrain/base](https://hub.docker.com/r/nextstrain/base/tags) and [nextstrain/base-builder](https://hub.docker.com/r/nextstrain/base-builder/tags)
- [x] Run zika-tutorial on linux/arm64 image "natively" using M1 Mac (54s run time)
- [x] Run zika-tutorial on linux/amd64 image natively using Linux on GitHub Codespace (50s run time)
- [x] Run ncov-tutorial on linux/arm64 image "natively" using M1 Mac (2m 23s run time)
- [x] Run ncov-tutorial on linux/amd64 image natively using Linux on GitHub Codespace (3m 6s run time)
- [x] ~test linux/arm64 image without emulation (all my M1 Macs have Rosetta enabled, and [it isn't easy to disable](https://developer.apple.com/forums/thread/669486?answerId=681889022#681889022).~ Not doing this since this feature is targeted towards other Apple silicon users which should also have Rosetta enabled.

### Summary of comment threads

- Local registry? ([ref](https://github.com/nextstrain/docker-base/pull/50#discussion_r955533198))
- Self-hosted runners to avoid emulation slowdowns? ([ref](https://github.com/nextstrain/docker-base/pull/50#discussion_r961056450))
- Cross-compilation? ([ref](https://github.com/nextstrain/docker-base/pull/50#discussion_r986331019))
- Emulation of pre-built binaries ([ref](https://github.com/nextstrain/docker-base/pull/50#discussion_r971207015))
- Compiling/runtime issues in `arm64` build
    - RAxML ([ref](https://github.com/nextstrain/docker-base/pull/50#discussion_r961092320))
    - matplotlib ([ref](https://github.com/nextstrain/docker-base/pull/50#discussion_r956218475))
    - Nextclade ([ref](https://github.com/nextstrain/docker-base/pull/50#discussion_r971105452))
    - cvxopt ([ref](https://github.com/nextstrain/docker-base/pull/50#discussion_r981811688))

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,8 +46,14 @@ jobs:
       tag: ${{ env.TAG }}
 
   test:
+    name: test (${{ matrix.platform }})
     needs: build
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        platform:
+          - linux/amd64
+          - linux/arm64
     steps:
 
     - uses: docker/login-action@v1
@@ -60,11 +66,18 @@ jobs:
       with:
         python-version: ~3
 
+    # The ubuntu-latest runner is linux/amd64 so anything else will
+    # run with emulation, which is still better than nothing.
+    - if: matrix.platform != 'linux/amd64'
+      uses: docker/setup-qemu-action@v2
+
     - name: Run zika-tutorial
       run: |
         git clone https://github.com/nextstrain/zika-tutorial
         pip install nextstrain-cli
         nextstrain build --image ghcr.io/nextstrain/base:${{ needs.build.outputs.tag }} zika-tutorial -F
+      env:
+        DOCKER_DEFAULT_PLATFORM: ${{ matrix.platform }}
 
   push-branch:
     if: startsWith(needs.build.outputs.tag, 'branch-') && github.event_name != 'pull_request'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,6 +30,8 @@ jobs:
       # and hyphens.
       run: echo "TAG=branch-${GITHUB_REF_NAME//[^A-Za-z0-9._-]/-}" | tee -a $GITHUB_ENV
 
+    - uses: docker/setup-qemu-action@v2
+
     # GITHUB_TOKEN is unreliable¹ so use a token from nextstrain-bot.
     # ¹ https://github.com/docker/build-push-action/issues/463#issuecomment-939394233
     - uses: docker/login-action@v1
@@ -38,7 +40,7 @@ jobs:
         username: nextstrain-bot
         password: ${{ secrets.GH_TOKEN_NEXTSTRAIN_BOT_MANAGE_PACKAGES }}
 
-    - run: ./devel/build -p linux/amd64 -r ghcr.io -t $TAG
+    - run: ./devel/build -p linux/amd64,linux/arm64 -r ghcr.io -t $TAG
 
     outputs:
       tag: ${{ env.TAG }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,7 +38,7 @@ jobs:
         username: nextstrain-bot
         password: ${{ secrets.GH_TOKEN_NEXTSTRAIN_BOT_MANAGE_PACKAGES }}
 
-    - run: ./devel/build -r ghcr.io -t $TAG
+    - run: ./devel/build -p linux/amd64 -r ghcr.io -t $TAG
 
     outputs:
       tag: ${{ env.TAG }}
--- a/Dockerfile
+++ b/Dockerfile
@@ -183,16 +183,12 @@ COPY builder-scripts/ /builder-scripts/
 
 # Download Nextalign v2
 # Set default Nextalign version to 2
-# NOTE: Running this program requires support for emulation on the Docker host
-# if the processor architecture is not amd64.
-RUN curl -fsSL -o /final/bin/nextalign2 https://github.com/nextstrain/nextclade/releases/latest/download/nextalign-x86_64-unknown-linux-gnu \
+RUN curl -fsSL -o /final/bin/nextalign2 https://github.com/nextstrain/nextclade/releases/latest/download/nextalign-$(/builder-scripts/target-triple) \
  && ln -sv nextalign2 /final/bin/nextalign
 
 # Download Nextclade v2
 # Set default Nextclade version to 2
-# NOTE: Running this program requires support for emulation on the Docker host
-# if the processor architecture is not amd64.
-RUN curl -fsSL -o /final/bin/nextclade2 https://github.com/nextstrain/nextclade/releases/latest/download/nextclade-x86_64-unknown-linux-gnu \
+RUN curl -fsSL -o /final/bin/nextclade2 https://github.com/nextstrain/nextclade/releases/latest/download/nextclade-$(/builder-scripts/target-triple) \
  && ln -sv nextclade2 /final/bin/nextclade
 
 # Fauna
--- a/Dockerfile
+++ b/Dockerfile
@@ -148,6 +148,27 @@ RUN curl -fsSL https://github.com/lh3/minimap2/releases/download/v2.24/minimap2-
 
 # 3. Install programs via pip
 
+# Build cvxopt on linux/arm64
+# cvxopt, an Augur dependency, does not have pre-built binaries for linux/arm64.
+#
+# First, add system deps for building¹:
+# - libopenblas-dev: Contains optimized versions of BLAS and LAPACK.
+# - libsuitesparse-dev: Contains SuiteSparse.
+#
+# Then, "install" (build) separately since the process requires a special
+# environment variable².
+#
+# ¹ https://cvxopt.org/install/#building-and-installing-from-source
+# ² https://github.com/cvxopt/cvxopt/issues/125#issuecomment-407396491
+RUN if [[ "$TARGETPLATFORM" == linux/arm64 ]]; then \
+      apt-get update && apt-get install -y --no-install-recommends \
+          libopenblas-dev \
+          libsuitesparse-dev \
+   && CVXOPT_SUITESPARSE_INC_DIR=/usr/include/suitesparse \
+      pip3 install cvxopt \
+      ; \
+    fi
+
 # Install envdir, which is used by pathogen builds
 RUN pip3 install envdir==1.0.1
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -89,6 +89,7 @@ RUN curl -fsSL https://github.com/vcftools/vcftools/releases/download/v0.1.16/vc
 # Download MAFFT
 # NOTE: Running this program requires support for emulation on the Docker host
 # if the processor architecture is not amd64.
+# TODO: Build from source to avoid emulation. Instructions: https://mafft.cbrc.jp/alignment/software/installation_without_root.html
 WORKDIR /download/mafft
 RUN curl -fsSL https://mafft.cbrc.jp/alignment/software/mafft-7.475-linux.tgz \
   | tar xzvpf - --no-same-owner --strip-components=2 mafft-linux64/mafftdir/ \
@@ -98,6 +99,7 @@ RUN curl -fsSL https://mafft.cbrc.jp/alignment/software/mafft-7.475-linux.tgz \
 # Download IQ-TREE
 # NOTE: Running this program requires support for emulation on the Docker host
 # if the processor architecture is not amd64.
+# TODO: Build from source to avoid emulation. Instructions: http://www.iqtree.org/doc/Compilation-Guide
 WORKDIR /download/IQ-TREE
 RUN curl -fsSL https://github.com/iqtree/iqtree2/releases/download/v2.1.2/iqtree-2.1.2-Linux.tar.gz \
   | tar xzvpf - --no-same-owner --strip-components=1 \
@@ -106,16 +108,19 @@ RUN curl -fsSL https://github.com/iqtree/iqtree2/releases/download/v2.1.2/iqtree
 # Download Nextalign v1
 # NOTE: Running this program requires support for emulation on the Docker host
 # if the processor architecture is not amd64.
+# TODO: Build from source to avoid emulation. Example: https://github.com/nextstrain/nextclade/blob/1.11.0/.circleci/config.yml#L183-L223
 RUN curl -fsSL -o /final/bin/nextalign1 https://github.com/nextstrain/nextclade/releases/download/1.11.0/nextalign-Linux-x86_64
 
 # Download Nextclade v1
 # NOTE: Running this program requires support for emulation on the Docker host
 # if the processor architecture is not amd64.
+# TODO: Build from source to avoid emulation. Example: https://github.com/nextstrain/nextclade/blob/1.11.0/.circleci/config.yml#L183-L223
 RUN curl -fsSL -o /final/bin/nextclade1 https://github.com/nextstrain/nextclade/releases/download/1.11.0/nextclade-Linux-x86_64
 
 # Download tsv-utils
 # NOTE: Running this program requires support for emulation on the Docker host
 # if the processor architecture is not amd64.
+# TODO: Build from source to avoid emulation. Instructions: https://github.com/eBay/tsv-utils/tree/v2.2.0#build-from-source-files
 RUN curl -L -o tsv-utils.tar.gz https://github.com/eBay/tsv-utils/releases/download/v2.2.0/tsv-utils-v2.2.0_linux-x86_64_ldc2.tar.gz \
  && tar -x --no-same-owner -v -C /final/bin -z --strip-components 2 --wildcards -f tsv-utils.tar.gz "*/bin/*" \
  && rm -f tsv-utils.tar.gz
@@ -129,12 +134,14 @@ RUN curl -L https://github.com/shenwei356/seqkit/releases/download/v2.2.0/seqkit
 # Download gofasta (for ncov/Pangolin)
 # NOTE: Running this program requires support for emulation on the Docker host
 # if the processor architecture is not amd64.
+# TODO: Build from source to avoid emulation. Instructions: https://github.com/virus-evolution/gofasta/tree/v0.0.6#installation
 RUN curl -fsSL https://github.com/virus-evolution/gofasta/releases/download/v0.0.6/gofasta-linux-amd64 \
   -o /final/bin/gofasta
 
 # Download minimap2 (for ncov/Pangolin)
 # NOTE: Running this program requires support for emulation on the Docker host
 # if the processor architecture is not amd64.
+# TODO: Build from source to avoid emulation. Instructions: https://github.com/lh3/minimap2/tree/v2.24#install
 RUN curl -fsSL https://github.com/lh3/minimap2/releases/download/v2.24/minimap2-2.24_x64-linux.tar.bz2 \
   | tar xjvpf - --no-same-owner --strip-components=1 -C /final/bin minimap2-2.24_x64-linux/minimap2
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -121,14 +121,10 @@ RUN curl -L -o tsv-utils.tar.gz https://github.com/eBay/tsv-utils/releases/downl
  && rm -f tsv-utils.tar.gz
 
 # Download csvtk
-# NOTE: Running this program requires support for emulation on the Docker host
-# if the processor architecture is not amd64.
-RUN curl -L https://github.com/shenwei356/csvtk/releases/download/v0.24.0/csvtk_linux_amd64.tar.gz | tar xz --no-same-owner -C /final/bin
+RUN curl -L https://github.com/shenwei356/csvtk/releases/download/v0.24.0/csvtk_${TARGETOS}_${TARGETARCH}.tar.gz | tar xz --no-same-owner -C /final/bin
 
 # Download seqkit
-# NOTE: Running this program requires support for emulation on the Docker host
-# if the processor architecture is not amd64.
-RUN curl -L https://github.com/shenwei356/seqkit/releases/download/v2.2.0/seqkit_linux_amd64.tar.gz | tar xz --no-same-owner -C /final/bin
+RUN curl -L https://github.com/shenwei356/seqkit/releases/download/v2.2.0/seqkit_${TARGETOS}_${TARGETARCH}.tar.gz | tar xz --no-same-owner -C /final/bin
 
 # Download gofasta (for ncov/Pangolin)
 # NOTE: Running this program requires support for emulation on the Docker host
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,7 +35,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 
 # Install a specific Node.js version
 # https://github.com/nodesource/distributions/blob/0d81da75/README.md#installation-instructions
-RUN curl -fsSL https://deb.nodesource.com/setup_16.x | bash - \
+RUN curl -fsSL https://deb.nodesource.com/setup_14.x | bash - \
  && apt-get update && apt-get install -y nodejs
 
 # Used for platform-specific instructions
@@ -297,7 +297,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 
 # Install a specific Node.js version
 # https://github.com/nodesource/distributions/blob/0d81da75/README.md#installation-instructions
-RUN curl -fsSL https://deb.nodesource.com/setup_16.x | bash - \
+RUN curl -fsSL https://deb.nodesource.com/setup_14.x | bash - \
  && apt-get update && apt-get install -y nodejs
 
 # Configure bash for interactive usage
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,6 +38,11 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 RUN curl -fsSL https://deb.nodesource.com/setup_16.x | bash - \
  && apt-get update && apt-get install -y nodejs
 
+# Used for platform-specific instructions
+ARG TARGETPLATFORM
+ARG TARGETOS
+ARG TARGETARCH
+
 # Add dependencies. All should be pinned to specific versions, except
 # Nextstrain-maintained software.
 # This includes pathogen-specific workflow dependencies. Since we only maintain a
--- a/Dockerfile
+++ b/Dockerfile
@@ -223,8 +223,8 @@ RUN /builder-scripts/download-repo https://github.com/nextstrain/auspice release
 RUN pip3 install evofr
 
 # Add NCBI Datasets command line tools for access to NCBI Datsets Virus Data Packages
-RUN curl -fsSL -o /final/bin/datasets https://ftp.ncbi.nlm.nih.gov/pub/datasets/command-line/v2/linux-amd64/datasets
-RUN curl -fsSL -o /final/bin/dataformat https://ftp.ncbi.nlm.nih.gov/pub/datasets/command-line/v2/linux-amd64/dataformat
+RUN curl -fsSL -o /final/bin/datasets https://ftp.ncbi.nlm.nih.gov/pub/datasets/command-line/v2/linux-${TARGETARCH}/datasets
+RUN curl -fsSL -o /final/bin/dataformat https://ftp.ncbi.nlm.nih.gov/pub/datasets/command-line/v2/linux-${TARGETARCH}/dataformat
 
 # ———————————————————————————————————————————————————————————————————— #
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -169,6 +169,15 @@ RUN if [[ "$TARGETPLATFORM" == linux/arm64 ]]; then \
       ; \
     fi
 
+# Install jaxlib on linux/arm64
+# jaxlib, an evofr dependency, does not have official pre-built binaries for
+# linux/arm64. A GitHub user has provided them in a fork repo.
+# https://github.com/google/jax/issues/7097#issuecomment-1110730040
+RUN if [[ "$TARGETPLATFORM" == linux/arm64 ]]; then \
+      pip3 install https://github.com/yoziru/jax/releases/download/jaxlib-v0.3.25/jaxlib-0.3.25-cp310-cp310-manylinux2014_aarch64.manylinux_2_17_aarch64.whl \
+      ; \
+    fi
+
 # Install envdir, which is used by pathogen builds
 RUN pip3 install envdir==1.0.1
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -53,12 +53,14 @@ RUN mkdir -p /final/bin /final/share /final/libexec
 # 1. Build programs from source
 
 # Build RAxML
-# AVX should be widely-supported enough
+# linux/arm64 does not support -mavx and -msse3 compilation flags which are used in the official repository.
+# Make these changes in a fork for now: https://github.com/nextstrain/standard-RAxML/tree/simde
+# TODO: Use the official repository if this PR is ever merged: https://github.com/stamatak/standard-RAxML/pull/50
 WORKDIR /build/RAxML
-RUN curl -fsSL https://api.github.com/repos/stamatak/standard-RAxML/tarball/v8.2.12 \
+RUN curl -fsSL https://api.github.com/repos/nextstrain/standard-RAxML/tarball/4621552064304a219ff03810f5f0d91e1063b68f \
   | tar xzvpf - --no-same-owner --strip-components=1 \
- && make -f Makefile.AVX.PTHREADS.gcc \
- && cp -p raxmlHPC-PTHREADS-AVX /final/bin
+  && make -f Makefile.AVX.PTHREADS.gcc \
+  && cp -p raxmlHPC-PTHREADS-AVX /final/bin
 
 # Build FastTree
 WORKDIR /build/FastTree
--- a/Dockerfile
+++ b/Dockerfile
@@ -87,6 +87,8 @@ RUN curl -fsSL https://github.com/vcftools/vcftools/releases/download/v0.1.16/vc
 # 2. Download pre-built programs
 
 # Download MAFFT
+# NOTE: Running this program requires support for emulation on the Docker host
+# if the processor architecture is not amd64.
 WORKDIR /download/mafft
 RUN curl -fsSL https://mafft.cbrc.jp/alignment/software/mafft-7.475-linux.tgz \
   | tar xzvpf - --no-same-owner --strip-components=2 mafft-linux64/mafftdir/ \
@@ -94,33 +96,49 @@ RUN curl -fsSL https://mafft.cbrc.jp/alignment/software/mafft-7.475-linux.tgz \
  && cp -p libexec/* /final/libexec
 
 # Download IQ-TREE
+# NOTE: Running this program requires support for emulation on the Docker host
+# if the processor architecture is not amd64.
 WORKDIR /download/IQ-TREE
 RUN curl -fsSL https://github.com/iqtree/iqtree2/releases/download/v2.1.2/iqtree-2.1.2-Linux.tar.gz \
   | tar xzvpf - --no-same-owner --strip-components=1 \
  && mv bin/iqtree2 /final/bin/iqtree
 
 # Download Nextalign v1
+# NOTE: Running this program requires support for emulation on the Docker host
+# if the processor architecture is not amd64.
 RUN curl -fsSL -o /final/bin/nextalign1 https://github.com/nextstrain/nextclade/releases/download/1.11.0/nextalign-Linux-x86_64
 
 # Download Nextclade v1
+# NOTE: Running this program requires support for emulation on the Docker host
+# if the processor architecture is not amd64.
 RUN curl -fsSL -o /final/bin/nextclade1 https://github.com/nextstrain/nextclade/releases/download/1.11.0/nextclade-Linux-x86_64
 
 # Download tsv-utils
+# NOTE: Running this program requires support for emulation on the Docker host
+# if the processor architecture is not amd64.
 RUN curl -L -o tsv-utils.tar.gz https://github.com/eBay/tsv-utils/releases/download/v2.2.0/tsv-utils-v2.2.0_linux-x86_64_ldc2.tar.gz \
  && tar -x --no-same-owner -v -C /final/bin -z --strip-components 2 --wildcards -f tsv-utils.tar.gz "*/bin/*" \
  && rm -f tsv-utils.tar.gz
 
 # Download csvtk
+# NOTE: Running this program requires support for emulation on the Docker host
+# if the processor architecture is not amd64.
 RUN curl -L https://github.com/shenwei356/csvtk/releases/download/v0.24.0/csvtk_linux_amd64.tar.gz | tar xz --no-same-owner -C /final/bin
 
 # Download seqkit
+# NOTE: Running this program requires support for emulation on the Docker host
+# if the processor architecture is not amd64.
 RUN curl -L https://github.com/shenwei356/seqkit/releases/download/v2.2.0/seqkit_linux_amd64.tar.gz | tar xz --no-same-owner -C /final/bin
 
 # Download gofasta (for ncov/Pangolin)
+# NOTE: Running this program requires support for emulation on the Docker host
+# if the processor architecture is not amd64.
 RUN curl -fsSL https://github.com/virus-evolution/gofasta/releases/download/v0.0.6/gofasta-linux-amd64 \
   -o /final/bin/gofasta
 
 # Download minimap2 (for ncov/Pangolin)
+# NOTE: Running this program requires support for emulation on the Docker host
+# if the processor architecture is not amd64.
 RUN curl -fsSL https://github.com/lh3/minimap2/releases/download/v2.24/minimap2-2.24_x64-linux.tar.bz2 \
   | tar xjvpf - --no-same-owner --strip-components=1 -C /final/bin minimap2-2.24_x64-linux/minimap2
 
@@ -169,11 +187,15 @@ COPY builder-scripts/ /builder-scripts/
 
 # Download Nextalign v2
 # Set default Nextalign version to 2
+# NOTE: Running this program requires support for emulation on the Docker host
+# if the processor architecture is not amd64.
 RUN curl -fsSL -o /final/bin/nextalign2 https://github.com/nextstrain/nextclade/releases/latest/download/nextalign-x86_64-unknown-linux-gnu \
  && ln -sv nextalign2 /final/bin/nextalign
 
 # Download Nextclade v2
 # Set default Nextclade version to 2
+# NOTE: Running this program requires support for emulation on the Docker host
+# if the processor architecture is not amd64.
 RUN curl -fsSL -o /final/bin/nextclade2 https://github.com/nextstrain/nextclade/releases/latest/download/nextclade-x86_64-unknown-linux-gnu \
  && ln -sv nextclade2 /final/bin/nextclade
 
--- a/README.md
+++ b/README.md
@@ -93,7 +93,8 @@ To add a software program to `nextstrain/base`, follow steps in this order:
 2. Check if it is available via PyPI. You can search on [PyPI's
    website](https://pypi.org/search/). If available, add an install command to
    the section labeled with `Install programs via pip`.
-3. Check if a pre-built binary is available on the software's website (e.g.
+3. Check if a pre-built binary for the `linux/amd64` platform (name contains
+   `linux` and `amd64`/`x86_64`) is available on the software's website (e.g.
    GitHub release assets). If available, add a download command to the section
    labeled with `Download pre-built programs`.
 4. The last resort is to build from source. Look for instructions on the
--- a/README.md
+++ b/README.md
@@ -103,9 +103,14 @@ To add a software program to `nextstrain/base`, follow steps in this order:
    `linux` and `amd64`/`x86_64`) is available on the software's website (e.g.
    GitHub release assets). If available, add a download command to the section
    labeled with `Download pre-built programs`.
+    - If a pre-built binary supporting `linux/arm64` (name contains `linux` and
+      `arm64`/`aarch64`) is also available, that should be used conditionally on
+      `ARG`s `TARGETPLATFORM` or `TARGETOS`+`TARGETARCH` in the Dockerfile. See
+      existing usage of those arguments for examples.
 4. The last resort is to build from source. Look for instructions on the
    software's website. Add a build command to the section labeled with `Build
-   programs from source`.
+   programs from source`. Note that this can require platform-specific
+   instructions.
 
 If possible, pin the software to a specific version. Otherwise, add the
 download/install/build command to the section labeled with `Add unpinned
--- a/README.md
+++ b/README.md
@@ -46,9 +46,15 @@ To build this image locally,
     ./devel/build
     ```
 
-    By default, this tags the image with `latest` and pushes to
+    By default, this builds for a single platform `linux/amd64`, tags the image with `latest`, and pushes to
     `localhost:5000`. See instructions at the top of the script for additional
     options.
+
+    If the target platform is different from the build platform, set up emulation before running `./devel/build`. This can be achieved using [`tonistiigi/binfmt`](https://github.com/tonistiigi/binfmt). For example, to set up emulation for `linux/arm64`, run:
+
+    ```
+    docker run --privileged --rm tonistiigi/binfmt --install arm64
+    ```
 
 On each subsequent change during your development iterations, you can run just
 the `./devel/build` command again.
--- a/devel/build
+++ b/devel/build
@@ -14,15 +14,17 @@
 set -euo pipefail
 
 # Set default values.
+platform=linux/amd64
 registry=localhost:5000
 tag=latest
 
 # Read command-line arguments.
-while getopts "r:t:" opt; do
+while getopts "p:r:t:" opt; do
     case "$opt" in
+        p) platform="$OPTARG";;
         r) registry="$OPTARG";;
         t) tag="$OPTARG";;
-        *) echo "Usage: $0 [-r <registry>] [-t <tag>]" 1>&2; exit 1;;
+        *) echo "Usage: $0 [-p <platform>] [-r <registry>] [-t <tag>]" 1>&2; exit 1;;
     esac
 done
 
@@ -35,8 +37,6 @@ export GIT_REVISION=$(git describe --tags --abbrev=40 --always --dirty || true)
 #
 # Calling `docker run nextstrain/base` will still only pull down the small base
 # image rather than pulling down the larger nextstrain/base-builder image.
-
-platform=linux/amd64
 
 # `buildx create` is necessary to use a driver that supports multi-platform
 # images.
--- a/builder-scripts/target-triple
+++ b/builder-scripts/target-triple
@@ -1,0 +1,25 @@
+#!/bin/bash
+#
+# Get the target triple from a Docker target platform.
+#
+# The Docker target platform is provided by the TARGETPLATFORM variable¹.
+# ¹ https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope
+#
+set -euo pipefail
+
+if [[ $# -gt 0 ]]; then
+    TARGETPLATFORM="$1"
+fi
+
+case $TARGETPLATFORM in
+    "linux/amd64")
+        echo "x86_64-unknown-linux-gnu"
+        ;;
+    "linux/arm64")
+        echo "aarch64-unknown-linux-gnu"
+        ;;
+    *)
+        echo "No target triple mapping for $TARGETPLATFORM." 1>&2
+        exit 1
+        ;;
+esac


victorlin:
Comparing to [the failure in #48 (using `build-push-action`)](https://github.com/nextstrain/docker-base/runs/6717402986?check_suite_focus=true), which points to:

https://github.com/nextstrain/docker-base/blob/163cba5590d615d089c0c881077eb5033a9929e0/Dockerfile#L39

and makes sense:

[The failure in this PR (using `docker buildx build`)](https://github.com/nextstrain/docker-base/runs/6890115833?check_suite_focus=true) points to:

https://github.com/nextstrain/docker-base/blob/163cba5590d615d089c0c881077eb5033a9929e0/Dockerfile#L13-L26

Not sure what's the reason for the discrepancy, but I also noticed that [the RAxML `make` command is one of the last things outputted in the logs](https://github.com/nextstrain/docker-base/runs/6890115833?check_suite_focus=true#step:5:2798).

Dockerfile
```diff
@@ -36,7 +36,8 @@ RUN curl -fsSL https://mafft.cbrc.jp/alignment/software/mafft-7.475-linux.tgz \
 WORKDIR /build/RAxML
 RUN curl -fsSL https://api.github.com/repos/stamatak/standard-RAxML/tarball/v8.2.12 \
   | tar xzvpf - --no-same-owner --strip-components=1
-RUN make -f Makefile.AVX.PTHREADS.gcc   # AVX should be widely-supported enough
+# AVX should be widely-supported enough
+RUN if [ "$TARGETPLATFORM" = "linux/amd64" ]; then make -f Makefile.AVX.PTHREADS.gcc; fi
```
victorlin, line 40: After adding 38a8fdcb692aa0d0366f6f33c7ee70b4ac0276c7 to skip compiling RAxML, `devel/build` ran for 25m before [failing on installing Augur dependencies](https://github.com/nextstrain/docker-base/runs/8027549975?check_suite_focus=true):

```
The following required packages can not be built:
freetype
Try installing freetype with `apt-get install
libfreetype6-dev` and pkg-config with `apt-get
install pkg-config`
```

devel/build
```diff
@@ -25,7 +25,7 @@ export GIT_REVISION=$(git describe --tags --abbrev=40 --always --dirty || true)
 # Calling `docker run nextstrain/base` will still only pull down the small base
 # image rather than pulling down the larger nextstrain/base-builder image.
 
-platform=linux/amd64
+platform=linux/amd64,linux/arm64
```
victorlin, line 28: This only works when emulation is configured properly (done for GitHub Actions via [`docker/setup-qemu-action`](https://github.com/docker/setup-qemu-action)). This means it won't be as portable to individual dev machines. Two options here:

1. Add emulation setup as a prerequisite to the build instructions in the README.
2. Make `devel/build` accept platforms as a parameter, and only use emulation when necessary.

I'm leaning towards (2) because I image there would be cases where one would want to build for just a single platform locally since it doesn't require emulation setup, and it will build much faster than a multi-platform build.

devel/build
```diff
@@ -47,18 +47,6 @@ docker buildx build \
     --cache-from $BASE_IMAGE \
     --cache-to type=inline \
     --tag $BASE_BUILDER_IMAGE:$tag \
-    --load \
+    --output type=oci,dest=test.tar \
```
victorlin, line 50: This is mostly an arbitrary value that supports multi-platform builds, retrieved from [the `--output` docs](https://docs.docker.com/engine/reference/commandline/buildx_build/#output). I'll refine it once the build is successful, along with trying [`docker manifest push`](https://docs.docker.com/engine/reference/commandline/manifest/#manifest-push).

victorlin (commented):
Made some progress today. Will investigate more another day.

Dockerfile
tsibley: The failure is inside the `pip` build of `matplotlib`, which is being built from source because `matplotlib` only provides pre-built arm64 wheels for macOS and not Linux. Building from source requires more development deps in the builder stage, e.g. `libfreetype6-dev` and maybe others ultimately. I believe it'll also require that the shared libraries themselves, e.g. `libfreetype6`, are installed in the final stage.

Stepping back a bit though, why does Augur still require `matplotlib`? I'd think that's a holdover from the early days.

tsibley: AFAICT, Augur only uses `matplotlib` (and `seaborn`) in

1. ad-hoc `scripts/`
2. optional code paths of `augur/titer_model.py` which are only used for ad-hoc dev/testing when running `python augur/titer_model.py` and are not used by `augur titers` (or any other command)
3. optional code paths of `augur/frequency_estimators.py` which are only used for ad-hoc dev/testing when running `python augur/frequency_estimators.py` and not used by `augur frequencies` (or any other command)

This suggests to me that of the three deps in Augur's `full` extra (`matplotlib`, `seaborn`, `cvxopt`), we should also provide `cvxopt` in a separate `titers` extra and then install the `titers` extra not the `full` extra in our Docker base image.

victorlin: https://github.com/nextstrain/augur/issues/672 😄

devel/build
tsibley: (2) sounds good to me.

Dockerfile
huddlej: Thanks! [I responded in the original issue](https://github.com/nextstrain/augur/issues/672#issuecomment-1233270756).

Dockerfile
```diff
@@ -36,7 +36,9 @@ RUN curl -fsSL https://mafft.cbrc.jp/alignment/software/mafft-7.475-linux.tgz \
...
```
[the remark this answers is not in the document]
victorlin: I noticed that [raxml_8.2.12+dfsg-4_**arm64**.deb](https://ubuntu.pkgs.org/20.04/ubuntu-universe-arm64/raxml_8.2.12+dfsg-4_arm64.deb.html) is available, which means it can be done.

Digging deeper on that page, the **Package Homepage** turns up a [link to a git repo](https://code.launchpad.net/~git-ubuntu-import/ubuntu/+source/raxml/+git/raxml):

```
git clone https://git.launchpad.net/~git-ubuntu-import/ubuntu/+source/raxml
```

The contents are similar to https://github.com/stamatak/standard-RAxML but importantly there is an additional [`debian/`](https://git.launchpad.net/~git-ubuntu-import/ubuntu/+source/raxml/tree/debian) directory.

Some lines in the file [`debian/rules`](https://git.launchpad.net/~git-ubuntu-import/ubuntu/+source/raxml/tree/debian/rules) are useful:

```make
ifeq (amd64,$(DEB_TARGET_ARCH_CPU))
##MFILES:=Makefile.gcc \ # we use a wrapper to the faster versions instead of shipping the "plain" version
MFILES =Makefile.PTHREADS.gcc \
	Makefile.SSE3.PTHREADS.gcc \
	Makefile.AVX.PTHREADS.gcc
MFLAGS=
else
MFILES = Makefile.AVX.PTHREADS.gcc
MFLAGS = CFLAGS="$(shell dpkg-buildflags --get CFLAGS) -DSIMDE_ENABLE_OPENMP -fopenmp-simd -O3"
endif
```

This means `Makefile.AVX.PTHREADS.gcc` is used to run on `arm64`, except with custom `CFLAGS` that override the [`CFLAGS` in the Makefile](https://github.com/stamatak/standard-RAxML/blob/a33ff40640b4a76abd5ea3a9e2f57b7dd8d854f6/Makefile.AVX.PTHREADS.gcc#L6) where [`-msse3` was causing our `arm64` compilation error](https://github.com/nextstrain/docker-base/pull/48#issuecomment-1154560485).

The file [`debian/patches/hardening`](https://git.launchpad.net/~git-ubuntu-import/ubuntu/+source/raxml/tree/debian/patches/hardening) is even more telling:

```diff
-CFLAGS := -D_USE_PTHREADS  -D__SIM_SSE3 -O2 -D_GNU_SOURCE -msse3 -fomit-frame-pointer -funroll-loops  -D__AVX #-Wall -pedantic -Wunused-parameter -Wredundant-decls  -Wreturn-type  -Wswitch-default -Wunused-value -Wimplicit  -Wimplicit-function-declaration  -Wimplicit-int -Wimport  -Wunused  -Wunused-function  -Wunused-label -Wno-int-to-pointer-cast -Wbad-function-cast  -Wmissing-declarations -Wmissing-prototypes  -Wnested-externs  -Wold-style-definition -Wstrict-prototypes  -Wdeclaration-after-statement -Wpointer-sign -Wextra -Wredundant-decls -Wunused -Wunused-function -Wunused-parameter -Wunused-value  -Wunused-variable -Wformat  -Wformat-nonliteral -Wparentheses -Wsequence-point -Wuninitialized -Wundef -Wbad-function-cast
+CFLAGS += -D_USE_PTHREADS  -D__SIM_SSE3 -O2 -D_GNU_SOURCE -fomit-frame-pointer -funroll-loops  -D__AVX #-Wall -pedantic -Wunused-parameter -Wredundant-decls  -Wreturn-type  -Wswitch-default -Wunused-value -Wimplicit  -Wimplicit-function-declaration  -Wimplicit-int -Wimport  -Wunused  -Wunused-function  -Wunused-label -Wno-int-to-pointer-cast -Wbad-function-cast  -Wmissing-declarations -Wmissing-prototypes  -Wnested-externs  -Wold-style-definition -Wstrict-prototypes  -Wdeclaration-after-statement -Wpointer-sign -Wextra -Wredundant-decls -Wunused -Wunused-function -Wunused-parameter -Wunused-value  -Wunused-variable -Wformat  -Wformat-nonliteral -Wparentheses -Wsequence-point -Wuninitialized -Wundef -Wbad-function-cast
 
+ifeq (amd64,$(shell dpkg-architecture -q DEB_TARGET_ARCH_CPU))
+	CFLAGS += -mavx -msse3
+endif
```

This patch updated the runner to only use `-msse3` on `amd64`, but the change isn't reflected in the Makefile that we download from GitHub which is why we still see the `-msse3` error.

I suspect that this should work for us using the same Makefile: `CFLAGS="-D_USE_PTHREADS -D__SIM_SSE3 -O2 -D_GNU_SOURCE -fomit-frame-pointer -funroll-loops -D__AVX"`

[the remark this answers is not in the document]
victorlin: It worked in a test run on 6ffbdbad142cfdb597466408e76a0769618f5ced (`devel/build` succeeded on the push-triggered run).

Added to this PR in d2cf41abcaa149cfd13871c67c5d773afd6892cc.

devel/build
victorlin: Done: 7efb3432b7b9aa5e8d12db229550058e985bebb5

Added an entry in testing checklist for manually checking that single-platform build still works locally.

devel/build
victorlin: Looking into this since build is successful now. However, I don't think it's as simple as getting the push to work. Based on some reading [[1]] [[2]], it doesn't seem possible to use the built multi-arch images locally, which is what we currently do for testing before pushing.

2 options I can think of:

1. Use one `docker buildx build` command per platform. This would allow us to continue our current pattern of build > test > tag > push. However, this would go against the purpose of using `buildx` in the first place, and I think we'd need different names for each platform which is undesirable.
2. Use `--push` here to push intermediate tags, which we then pull for testing (on each platform?) before pushing with the final `$TAG` (i.e. build > tag (intermediate) > push > pull > test > tag > push). This might be best solved by splitting CI into separate GitHub Actions workflows.

I think option 2 is more appropriate.

[1]: https://github.com/docker/buildx/issues/166#issuecomment-544827163
[2]: https://developer.arm.com/documentation/102475/0100/Multi-architecture-images

Dockerfile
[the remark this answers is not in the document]
victorlin: Compilation with those flags fail. I tried using both the existing reference to https://api.github.com/repos/stamatak/standard-RAxML/tarball/v8.2.12 and the Ubuntu git repo at https://git.launchpad.net/~git-ubuntu-import/ubuntu/+source/raxml. [Both fail with](https://github.com/nextstrain/docker-base/runs/8147469773?check_suite_focus=true#step:8:4116):

```
[linux/arm64 builder 6/6] RUN if [ "linux/arm64" = "linux/arm64" ]; then make -f Makefile.AVX.PTHREADS.gcc CFLAGS="-D_USE_PTHREADS  -D__SIM_SSE3 -O2 -D_GNU_SOURCE -fomit-frame-pointer -funroll-loops  -D__AVX"; fi
rm -f *.o raxmlHPC-PTHREADS-AVX
gcc  -D_USE_PTHREADS  -D__SIM_SSE3 -O2 -D_GNU_SOURCE -fomit-frame-pointer -funroll-loops  -D__AVX   -c -o axml.o axml.c
axml.c:70:10: fatal error: xmmintrin.h: No such file or directory
 #include <xmmintrin.h>
          ^~~~~~~~~~~~~
compilation terminated.
make: *** [<builtin>: axml.o] Error 1
```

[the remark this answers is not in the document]
victorlin: Ok, using the flags alone wasn't enough. There's actually a GitHub PR created by the patch author for similar intents, and it works after some additional changes: https://github.com/stamatak/standard-RAxML/pull/50#issuecomment-1235769882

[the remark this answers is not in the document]
victorlin: [`linux/arm64` RAxML build is now successful](https://github.com/nextstrain/docker-base/runs/8161756817?check_suite_focus=true#step:8:3818).

devel/build
tsibley: We could run a [local registry](https://hub.docker.com/_/registry) as a [sidecar service](https://docs.github.com/en/actions/using-containerized-services/about-service-containers) in CI and `--push` / `--output type=image,push=true` to that. Then we can do whatever testing we want and finally push the whole thing up to the public registry for the release phase of CI.

Dockerfile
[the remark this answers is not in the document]
tsibley: Oof. I guess it really is [as painfully slow as we were warned](https://github.com/nextstrain/docker-base/issues/35#issuecomment-1204318267).

If we're going to make this not slow by building arm64 without emulation, then we'd need to provision our own runner on our "own" hardware (e.g. rented from AWS or even a purchased M1/M2 Mac Mini) and use it in CI here. That has implications for [how the final multi-arch image is produced](https://github.com/nextstrain/docker-base/pull/50#discussion_r955533198). Since we'd no longer have the option of doing it all in one go, we'd have to upload the intermediate OCI tarballs as job artifacts from each native host (e.g. x86_64 and arm64) and then combine them afterwards with `docker manifest` in another job (e.g. like Nextstrain CLI arranges for its standalone builds).

devel/build
victorlin: @tsibley thanks, I hadn't thought of that nor used local registries before. I wonder what's the benefit of that vs. (2) above – is it just the lack of intermediate tags pushed to Docker Hub?

My main concern with (2) that also applies to your option is the expansion of `devel/build` to push to a registry, which historically has been the duty of `devel/push`. This would lead to usability of those scripts being tied to logic in the CI workflow (e.g. we'd have to set up a local registry for local development too?).

tsibley: Lack of public version of those tags, yeah. This is more important than it looks at first blush, I think, because I think all of those tags will need to be unique to a CI run to avoid collisions, so it won't just be N more mutable tags for N platforms (e.g. not just `nextstrain/base:arm64`) but an ongoing series of them (N tags × each CI run, e.g. `nextstrain/base:arm64-${uuid}`).

Agreed it sort of muddies the current clarity of separated concerns with the build steps, though I'd still see a separation where `devel/build` produces the (multi-arch) image by pushing to the local registry and `devel/push` is responsible for pushing it publicly. And yes, local dev would have to have a local registry running too (though we could make this ~easy if desired).

An alternative to the local registry would be for `devel/build` to export with `type=oci` (or similar) as you have already and then _also_ be responsible for loading that into the local Docker daemon so it's immediately usable after `devel/build`. Note that the image loaded into the local Docker daemon in this case **wouldn't** be suitable for pushing publicly, as it wouldn't be a multi-arch manifest. We'd still have to have a separate step that used `docker manifest create` on the OCI tarballs to push to the public registry.

tsibley: The advantage I see of the local registry over the alternative above is that the registry more closely matches (and thus tests) what an actual interaction looks like with the public registry and the built image. e.g. if we produce a bad multi-arch image, we're unlikely to catch it unless we're interacting with a registry in CI.

Dockerfile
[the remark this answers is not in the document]
tsibley: Ah, one further note about this idea of a Mac Mini as a build machine (mentioned above and [earlier in Slack](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1654885748224469?thread_ts=1654883833.977699&cid=C01LCTT7JNN)): currently the GitHub Actions runner [doesn't support arm64 on macOS](https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners#architectures), so that's moot for now (though I suspect it's a popularly requested feature).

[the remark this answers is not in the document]
tsibley: And it may be possible to enable dev/experimental support for arm64 on macOS, c.f. https://www.pantsbuild.org/docs/ci-for-macos-on-arm64#setting-up-the-self-hosted-runner.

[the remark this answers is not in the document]
tsibley: Another note in that vein: for our arm64 image builds here, we don't actually need macOS on arm64 hardware as we'd be running Linux in a VM on it in that case anyway. So Linux on arm64 hardware is actually preferable and these are available for rent from many sources as previously noted.

But for _testing_ our arm64 images on macOS in CI, we would need macOS running on arm64 hardware. We would also need that for building macOS arm64 standalone binaries for the CLI too.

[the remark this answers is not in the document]
victorlin: > though I suspect it's a popularly requested feature

Looks about right: https://github.com/actions/runner-images/issues/2187

devel/build
victorlin: > I think all of those tags will need to be unique to a CI run to avoid collisions

Why wouldn't this apply to the current implementation? The tag is the same for all runs on a PR, so it represents the latest run with a successful push:

https://github.com/nextstrain/docker-base/blob/c3857f55612e4ccc5834348ea51a9c57acb7f9f7/.github/workflows/ci.yml#L23

So I'd expect something like `TAG=branch-${GITHUB_REF_NAME//[^A-Za-z0-9._-]/-}-$ARCH` to behave similarly.

Regardless, I'm now experimenting with using a local registry as you described.

victorlin: These changes make use of a local registry: ecac84094b3effcac91bd098f69456f79a66fd32...9ca46fb98f441d645e99004d71521b4e7929ddde

.github/workflows/ci.yml
```diff
@@ -22,7 +22,25 @@ jobs:
...
+        source /etc/os-release
+        sudo sh -c "echo 'deb ${REPO_URL}/x${NAME}_${VERSION_ID}/ /' > /etc/apt/sources.list.d/devel:kubic:libcontainers:unstable.list"
+        sudo wget -qnv https://download.opensuse.org/repositories/devel:kubic:libcontainers:stable/x${NAME}_${VERSION_ID}/Release.key -O Release.key
+        sudo apt-key add Release.key
+        sudo apt-get update
+        sudo apt-get install skopeo
+        skopeo --version
+
+    - run: ./devel/start-localhost-registry
+
```
victorlin, line 42: This should have a teardown script paired with it.

devel/build-to-localhost
```diff
@@ -46,8 +47,8 @@ docker buildx build \
     --cache-from $BASE_BUILDER_IMAGE \
     --cache-from $BASE_IMAGE \
     --cache-to type=inline \
-    --tag $BASE_BUILDER_IMAGE:$tag \
-    --load \
+    --tag localhost:5000/$BASE_BUILDER_IMAGE:$tag \
+    --push \
```
victorlin, line 51: `localhost:5000` or at least the port `5000` should be parameterized for customization across these scripts. That would make 3 parameters to this script, which is a bit messy if they're arbitrarily positional. Maybe it'd be good to consider `getopt` for named parameters.

devel/localhost-to-dockerhub
```diff
@@ -0,0 +1,28 @@
...
+
+BASE_IMAGE="nextstrain/base"
+BASE_BUILDER_IMAGE="nextstrain/base-builder"
+
+# copy local $tag to remote $tag
+skopeo copy --multi-arch=all --src-tls-verify=false docker://localhost:5000/$BASE_IMAGE:$tag docker://docker.io/$BASE_IMAGE:$tag
+skopeo copy --multi-arch=all --src-tls-verify=false docker://localhost:5000/$BASE_BUILDER_IMAGE:$tag docker://docker.io/$BASE_BUILDER_IMAGE:$tag
+
+if [ $# -eq 2 ] && [$2 == "latest" ]; then
+    # copy local $tag to remote latest
```
victorlin, line 25: Probably also condition that `$1 != "latest"` so it isn't copied twice (probably wouldn't hurt, but also not necessary).

devel/build-to-localhost
victorlin, line 51: Since this means the built images are no longer available for use locally, maybe a script to call `docker pull localhost:5000/nextstrain/base:$tag` similar to the existing [`./devel/pull`](https://github.com/nextstrain/docker-base/blob/c3857f55612e4ccc5834348ea51a9c57acb7f9f7/devel/pull) would be reasonable to add alongside these changes.

Dockerfile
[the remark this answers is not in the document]
victorlin: Back to the main issue here (`arm64` emulation slowness).

> Linux on arm64 hardware is actually preferable and these are available for rent from many sources as previously noted.

This looks like a good guide detailing how to deploy/use arm64 self-hosted runners on AWS: https://noise.getoto.net/2021/11/01/building-arm64-applications-on-aws-graviton2-using-the-aws-cdk-and-self-hosted-runners-for-github-actions/

.github/workflows/ci.yml
```diff
@@ -41,10 +59,9 @@ jobs:
       run: |
         git clone https://github.com/nextstrain/zika-tutorial
         pip install nextstrain-cli
-        nextstrain build --image nextstrain/base:$TAG zika-tutorial -F
+        nextstrain build --image localhost:5000/nextstrain/base:$TAG zika-tutorial -F
```
victorlin, line 62: Might be a good idea to run this test with `--platform linux/arm64`, though it'll be using emulation for the time being.

.github/workflows/ci.yml
```diff
@@ -41,10 +59,9 @@ jobs:
...
```
victorlin, line 63: Now that CI succeeded and pushed [an arm64 image](https://hub.docker.com/layers/nextstrain/base/branch-victorlin-add-arm64-keep-scripts/images/sha256-352393ad341db327996d785778b01c4094c4555344225d09409cca422516e1ea), I pulled it locally on my M1 Mac for testing. It runs zika-tutorial fine but this error comes up when calling `nextalign` in the ncov workflow:

```
qemu-x86_64: Could not open '/lib64/ld-linux-x86-64.so.2': No such file or directory
```

I need to revisit all the `x86`/`amd64` references in the Dockerfile and either (1) build them from source or (2) set up emulation within the Docker image.

victorlin: All `x86`/`amd64` references in [Dockerfile](https://github.com/nextstrain/docker-base/blob/6f97b669959dbb123cf58321386fc4492087ffe0/Dockerfile):

- `nextalign2`, `nextclade2`: calling the command alone results in error above.
- `nextalign1`, `nextclade1`: calling the command alone runs fine.
- `csvtk`: the [dim/nrow/ncol examples](https://bioinf.shenwei.me/csvtk/usage/#dimnrowncol) run fine.
- `tsv-utils`: the [`tsv-pretty` example](https://github.com/eBay/tsv-utils/blob/38ed0a1c31742bd8b59196517e89ff0b51e8fb80/README.md#tsv-pretty) runs fine.
- `gofasta`: untested
- `seqkit`: untested

My thought now is that this is specific to `nextclade2`/`nextalign2`.

victorlin: 401506ed8d8dacf40f515d8ae9b4d45ef2b025ff should fix the error above, but still curious about emulation for the others.

devel/localhost-to-dockerhub
```diff
@@ -0,0 +1,28 @@
...
+fi
+
+tag=$1
+
+BASE_IMAGE="nextstrain/base"
+BASE_BUILDER_IMAGE="nextstrain/base-builder"
+
+# copy local $tag to remote $tag
+skopeo copy --multi-arch=all --src-tls-verify=false docker://localhost:5000/$BASE_IMAGE:$tag docker://docker.io/$BASE_IMAGE:$tag
+skopeo copy --multi-arch=all --src-tls-verify=false docker://localhost:5000/$BASE_BUILDER_IMAGE:$tag docker://docker.io/$BASE_BUILDER_IMAGE:$tag
```
victorlin, line 22: This [errored](https://github.com/nextstrain/docker-base/actions/runs/3055446154/jobs/4928511381), likely transient:

> level=fatal msg="copying image 2/2 from manifest list: trying to reuse blob sha256:828da3c2b6e5dfebc31218a341a7d947edf0fd0c700eb6241afe12ffc1ad11a1 at destination: Head \"https://production.cloudflare.docker.com/registry-v2/docker/registry/v2/blobs/sha256/82/828da3c2b6e5dfebc31218a341a7d947edf0fd0c700eb6241afe12ffc1ad11a1/data?verify=1663189957-ErKuGjE6SA66GsXxUQk9uagRVyw%3D\": net/http: TLS handshake timeout"

Seems like a network error on either [Docker](https://status.docker.com/) or [GitHub](https://www.githubstatus.com/history)'s end. Looking at the linked status pages, I don't see anything documented recently that would have caused this. Just noting here for reference.

devel/build
tsibley: > Why wouldn't this apply to the current implementation? The tag is the same for all runs on a PR, so it represents the latest run with a successful push…

Hmm, yes, but I _think_ it's subtly different because instead of all CI runs for a branch sharing a single tag, they're now sharing a _set of tags_. A single tag is simple "last one to push wins" behaviour, but a set of tags will at times be a mix of pushes from different CI runs. This is maybe acceptable if pushes are the final step and all subsequent pulls are by humans who can make judgement calls (e.g. the CI run I cared about pushed and I see there aren't any other CI runs). But with each CI run also _pulling_ what it previously pushed in order to make the multi-arch image, the potential for a run to get an image from another concurrent CI run seems like a recipe for unhappiness.

Dockerfile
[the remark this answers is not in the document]
tsibley: > However, it's probably reasonable to use emulation for a bit and [stay hopeful](https://twitter.com/iximiuz/status/1565629586880270336) for GitHub-hosted arm64 runners.

Agreed.

> This looks like a good guide detailing how to deploy/use arm64 self-hosted runners on AWS: https://noise.getoto.net/2021/11/01/building-arm64-applications-on-aws-graviton2-using-the-aws-cdk-and-self-hosted-runners-for-github-actions/

Nod. Like most how-to guides of this style, I would suggest using it as a rough list of steps to consider but not to actually copy/paste the examples nor treat it as a comprehensive (or even good) way to set the thing in question up. An artifact of a specific person's choices best used to guide your own understanding and support your own learning of how to set up the thing for your needs with your own choices.

devel/build
victorlin: I see, that makes sense. Though that doesn't apply currently since CI is only pulling `latest`, it would apply after #57 which is something I plan to look into soon.

devel/localhost-to-dockerhub
victorlin: Retry succeeded.

Dockerfile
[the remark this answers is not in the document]
tsibley: I wrote down [some of my thoughts on self-hosted runners](https://github.com/tsibley/blab-standup/blob/master/2022-09-14.md#readme).

.github/workflows/ci.yml
```diff
@@ -22,7 +22,25 @@ jobs:
...
+        sudo apt purge golang-github-containers-common skopeo
+        sudo apt autoremove --purge
+        REPO_URL="https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/unstable"
+        source /etc/os-release
+        sudo sh -c "echo 'deb ${REPO_URL}/x${NAME}_${VERSION_ID}/ /' > /etc/apt/sources.list.d/devel:kubic:libcontainers:unstable.list"
+        sudo wget -qnv https://download.opensuse.org/repositories/devel:kubic:libcontainers:stable/x${NAME}_${VERSION_ID}/Release.key -O Release.key
+        sudo apt-key add Release.key
+        sudo apt-get update
+        sudo apt-get install skopeo
+        skopeo --version
```
tsibley, line 39: This is a lot. Can we run a [containerized Skopeo](https://github.com/containers/skopeo/blob/main/install.md#container-images) instead? e.g.

```
docker run --rm -it quay.io/skopeo/stable --version
```

Dockerfile
```diff
@@ -164,16 +166,24 @@ COPY --from=builder \
...
 
+ARG TARGETPLATFORM
+
+# use /env to persist variable across RUNs
+RUN if [ "$TARGETPLATFORM" = "linux/amd64" ]; then echo "export NEXTCLADE_PLATFORM=x86_64-unknown-linux-gnu" >> /env; fi
+RUN if [ "$TARGETPLATFORM" = "linux/arm64" ]; then echo "export NEXTCLADE_PLATFORM=aarch64-unknown-linux-gnu" >> /env; fi
+
 # Add Nextalign v2
-RUN curl -fsSL https://github.com/nextstrain/nextclade/releases/latest/download/nextalign-x86_64-unknown-linux-gnu \
+RUN . /env; curl -fsSL https://github.com/nextstrain/nextclade/releases/latest/download/nextalign-$NEXTCLADE_PLATFORM \
```
tsibley, line 176: There's also `TARGETARCH` [available](https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope), and `ARG`s are available in the environment of a `RUN` (which is [not subject to variable substitution](https://docs.docker.com/engine/reference/builder/#environment-replacement) at the Dockerfile level), so an alternative to this unusual sort of roundabout `./env` thing is something like:

```Dockerfile
ARG TARGETARCH
RUN curl …/download/nextalign-$(case $TARGETARCH in arm64) echo aarch64;; amd64) echo x86_64;; esac)-unknown-linux-gnu
```

The mapping of Docker-style archs to standard/Rust-style archs is an unfortunate necessity.

Dockerfile
```diff
@@ -164,16 +166,24 @@ COPY --from=builder \
...
       --output /usr/local/bin/nextalign2 \
  && chmod a+rx /usr/local/bin/nextalign2
 
 # Add Nextclade v2
-RUN curl -fsSL https://github.com/nextstrain/nextclade/releases/latest/download/nextclade-x86_64-unknown-linux-gnu \
+RUN . /env; curl -fsSL https://github.com/nextstrain/nextclade/releases/latest/download/nextclade-$NEXTCLADE_PLATFORM \
       --output /usr/local/bin/nextclade2 \
  && chmod a+rx /usr/local/bin/nextclade2
 
+RUN rm /env
```
tsibley, line 185: This results in `/env` not being present in the final image's final layer, but additional layers are still added to the final image for this `RUN` (and the two `RUN`s that create `/env`).

The number of layers matters even if the layer itself is tiny. The way to avoid inflating number of layers is to move this conditional build-only stuff into the build stage and then copy the binaries from there.

(Although maybe buildx can do fancier things here these days…?)

devel/localhost-to-dockerhub
tsibley: Could be reduced to `[[ "$tag" != latest && "${2:-}" == latest ]]`.

Better to use `[[ … ]]` expressions, regardless.

devel/build
tsibley: Yes, but I think it would apply to using `docker manifest create` to assemble the per-arch images into a multi-arch image outside of `docker buildx build`. AIUI, `docker manifest create` would be querying the registry for the info of the per-arch images, even if it's not downloading their layers.

devel/localhost-to-dockerhub
victorlin: Nice, updated: 59c99bf3a49275548a40bc5aa5df0f92e5686a8c

.github/workflows/ci.yml
victorlin: Good idea, done: 1769212f6fa8b31e86da07cbe6391c87cfc9c44a

tsibley: Might need adjustment to pass in credentials for docker.io to the skopeo container?

Dockerfile
victorlin: I implemented another alternative: 70443f60fa6d33f9e16780db6ca5dbbd491f3a1b

This puts the conditional logic in a script, which is then invoked inline to construct the URL.

Also, I chose `TARGETPLATFORM` over `TARGETARCH` since I think it's nice to include the OS component in the mapping when it's included in the URLs as well. This also keeps the door open for other OS builds, e.g.

- `linux/amd64`-> `x86_64-unknown-linux-gnu`
- `linux/arm64` -> `aarch64-unknown-linux-gnu`
- `windows/amd64` -> `x86_64-pc-windows-gnu.exe`

Dockerfile
victorlin: In 70443f60fa6d33f9e16780db6ca5dbbd491f3a1b I got rid of the extra `/env` layers, and moved the commands up to the builder stage to access the `devel/` script.

> maybe buildx can do fancier things here these days…?

From my searching yesterday, everything I found still suggested workarounds for not having access to environment variables between `RUN` commands. 😕

.github/workflows/ci.yml
victorlin: Oh that makes sense... probably need to pass `secrets.DOCKER_LOGIN` and `secrets.DOCKER_PASSWORD` when calling the script:

https://github.com/nextstrain/docker-base/blob/70443f60fa6d33f9e16780db6ca5dbbd491f3a1b/.github/workflows/ci.yml#L53

I think 4 arguments calls for using named args with `getopt`, or expect them to be available as env vars.

tsibley: Yeah, they should be env vars, and I expect they should be passable to skopeo's container as env vars too.

tsibley: Ah, looks like it doesn't support passing via env vars ([ref](https://github.com/containers/skopeo/blob/main/docs/skopeo-copy.1.md)), but you can have it read the Docker auth file by mounting it into the container with:

```
-v $HOME/.docker/config.json:/tmp/auth.json:ro
```

or a bit more scoped:

```
-v $HOME/.docker/config.json:/docker-auth.json:ro … --dest-authfile /docker-auth.json
```

Dockerfile
tsibley: > From my searching yesterday, everything I found still suggested workarounds for not having access to environment variables between `RUN` commands. confused

Ah, I meant in terms of pruning where layer A adds file X and layer D removes file X.

.github/workflows/ci.yml
victorlin: Done in f83877f6320b22bb4d6e5c449fdb4d89c73990a1:

https://github.com/nextstrain/docker-base/blob/f83877f6320b22bb4d6e5c449fdb4d89c73990a1/devel/localhost-to-dockerhub#L30-L43

.github/workflows/ci.yml
victorlin: Added in acbde816baf00f147e1d455e4691257684e5e21f:

https://github.com/nextstrain/docker-base/blob/acbde816baf00f147e1d455e4691257684e5e21f/devel/stop-localhost-registry#L3-L9

devel/build-to-localhost
victorlin: Also noting that I don't think `./devel/pull` is useful in CI, though it's still good to keep around for local dev. With single-platform builds it is redundant: `docker buildx build --cache-from` should pull if it doesn't exist locally yet. With multi-platform builds it is redundant and incomplete: unless specifying `--platform` and retagging to avoid overwrite, `pull` will not keep layers for all platforms locally.

devel/build
victorlin: I think this is resolved. Local registry is working pretty well in the latest iteration (fcc2b72f18b5ddb243dfff57c27dbc87fbfc88be...ff2edf1190bb2ce9a0323e84992a4111bb0963d7).

We don't have to mess with `docker manifest` in this approach.

.github/workflows/ci.yml
```diff
@@ -22,7 +22,11 @@ jobs:
       # and hyphens.
       run: echo "TAG=branch-${GITHUB_REF_NAME//[^A-Za-z0-9._-]/-}" | tee -a $GITHUB_ENV
 
-    - run: ./devel/build $TAG
+    - uses: docker/setup-qemu-action@v2
+
+    - run: ./devel/start-localhost-registry
+
+    - run: ./devel/build-to-localhost $TAG linux/amd64,linux/arm64
```
victorlin, line 30: With the release of [Augur 18.0.0](https://github.com/nextstrain/augur/releases/tag/18.0.0) which makes cvxopt a required dependency, another dependency of that [fails to compile](https://github.com/nextstrain/docker-base/actions/runs/3139336657/jobs/5099809084):

```
gcc -pthread -shared -Wl,--strip-all build/temp.linux-aarch64-3.7/src/C/base.o build/temp.linux-aarch64-3.7/src/C/dense.o build/temp.linux-aarch64-3.7/src/C/sparse.o -L/usr/lib -L/usr/local/lib -lm -llapack -lblas -lpython3.7m -o build/lib.linux-aarch64-3.7/cvxopt/base.cpython-37m-aarch64-linux-gnu.so
/usr/bin/ld: cannot find -llapack
/usr/bin/ld: cannot find -lblas
collect2: error: ld returned 1 exit status
error: command 'gcc' failed with exit status 1
```

This is similar to https://github.com/nextstrain/docker-base/pull/50#discussion_r956218475, but this time we can't ignore the dependency. Options:

1. Install whatever is necessary to compile properly (maybe `sudo apt-get install liblapack-dev libblas-dev`?)
2. Install pre-built `linux-aarch64` binaries from [conda-forge](https://anaconda.org/conda-forge/cvxopt).

I'm planning on (1) since (2) seems like a bigger change to the Dockerfile.

devel/build-to-localhost
tsibley: Nod. What's the use case for it in local dev then?

devel/build
tsibley: Awesome!

.github/workflows/ci.yml
tsibley: (1) or its equivalent seems best, yes. There are several Ubuntu packages providing libblas.so and liblapack.so, which should roughly match what Debian in the container image here has too.

image

In prior versions of this image (back on Alpine, so packaging names are a bit different), we compiled cvxopt like so:

https://github.com/nextstrain/docker-base/blob/3917c1b9aa0aa19e694f94f6bd3e56418f248269/Dockerfile#L89-L94

This used openblas (Ubuntu package `libopenblas-dev` above) to provide libblas.so and liblapack.so and also installed libsuitesparse for use by cvxopt (which itself relies on blas and lapack).

tsibley: It's too bad that cvxopt isn't [using cibuildwheels for multi-arch wheels](https://cibuildwheel.readthedocs.io/en/stable/options/#archs).

devel/build-to-localhost
victorlin: Hm, actually probably not useful for local dev either. I previously thought that `--cache-from` would not pull from the registry if an image with the same tag is available locally, but [the docs](https://docs.docker.com/engine/reference/commandline/buildx_build/#cache-from) say otherwise:

> Use an external cache source for a build (--cache-from)
> If no type is specified, `registry` exporter is used with a specified reference.

So I think we can just get rid of `./devel/pull` independent from the changes in this PR.

tsibley: Sounds good.

victorlin: #87

.github/workflows/ci.yml
victorlin: Interesting historical info!

2bc2fa1bf13f12174a667982f40ff63d710fad86 should fix this.

devel/build-to-localhost
victorlin: > maybe a script to call `docker pull localhost:5000/nextstrain/base:$tag` similar to the existing [`./devel/pull`](https://github.com/nextstrain/docker-base/blob/c3857f55612e4ccc5834348ea51a9c57acb7f9f7/devel/pull) would be reasonable to add alongside these changes.

Done with 31ef757733c939ac7d801cbd81d628f4aaf38993.

Dockerfile
```diff
@@ -28,6 +28,27 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
+fi
+
+# "Install" (build) cvxopt separately since the process requires a special
+# environment variable.
+# https://github.com/cvxopt/cvxopt/issues/125#issuecomment-407396491
+RUN if [ "$TARGETPLATFORM" = "linux/arm64" ]; then \
+  CVXOPT_SUITESPARSE_INC_DIR=/usr/include/suitesparse \
+  pip3 install cvxopt \
+  ; \
+fi
```
tsibley, line 50: Maybe combine this into the previous RUN to avoid the two extra useless layers for non-arm64 images?

Dockerfile
```diff
@@ -28,6 +28,27 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
 ARG TARGETPLATFORM
 
+# cvxopt, an Augur dependency, does not have pre-built binaries for linux/arm64,
+# so it will have to be built on that platform.
+
+# Add system deps for building cvxopt
+RUN if [ "$TARGETPLATFORM" = "linux/arm64" ]; then \
+  apt-get install -y --no-install-recommends \
+      libblas-dev \
+      liblapack-dev \
```
tsibley, line 38: libblas-dev and liblapack-dev are the unoptimized reference implementations from https://netlib.org. We probably want to stick with OpenBLAS (libopenblas-dev) or [another optimized implementation of them](https://netlib.org/blas/faq.html#_5_a_id_are_optimized_blas_libraries_available_where_can_i_find_vendor_supplied_blas_a_are_optimized_blas_libraries_available_where_can_i_find_optimized_blas_libraries). OpenBLAS is what cvxopt compiles its own wheels with.

Dockerfile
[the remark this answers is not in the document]
victorlin: A better solution for speeding up build time is **cross-compilation** mentioned in [the buildkit docs](https://github.com/moby/buildkit/blob/f2c41e15cb4d725649efe47a8d3023d0b35fdbe8/docs/multi-platform.md#builds-are-very-slow-through-emulation) and in [this guide](https://medium.com/@tonistiigi/faster-multi-platform-builds-dockerfile-cross-compilation-guide-part-1-ec087c719eaf). Basically, use `FROM --platform=$BUILDPLATFORM` to avoid emulation. I'll give it a shot in this PR.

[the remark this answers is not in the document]
tsibley: Ah, nice. We'd still need emulation (or self-hosted runners) for testing though, yeah?

(And I think also need them for other contexts/purposes where we can't cross-compile, but unrelated to this image.)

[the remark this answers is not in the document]
victorlin: Yes, definitely!

Dockerfile
```diff
@@ -25,6 +25,30 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         npm \
         pkg-config
 
+# Used for platform-specific instructions
+ARG TARGETPLATFORM
```
victorlin, line 29: Starting a new "thread" on cross-compilation.

Looked into this a bit today, and it's not so simple given the current Dockerfile. `FROM --platform=$BUILDPLATFORM` is only half of the equation - the other half is getting the output to be `TARGETPLATFORM`-native. [tonistiigi/xx](https://github.com/tonistiigi/xx) provides drop-in replacements for build tools like `gcc`, so what's covered by that is not a problem.

The problem is with various package managers not supported by `xx` (e.g. `pip` and `npm`). As far as I can tell, `pip` will install based on the caller's platform. So if using `FROM --platform=$BUILDPLATFORM`, this will not reflect the `TARGETPLATFORM`.

One idea is to have two builder stages:

1. `builder` to compile using `xx` (no emulation, much faster)
2. `builder-emulation` to install from package managers (slower, but dunno how else to do this).

This also isn't ideal because some Python packages have dependencies that don't provide pre-compiled arm64 wheels and thus requires building, which defeats the point of cross-compilation in the first place. We could manually build all these in the non-emulation stage (like 2bc2fa1bf13f12174a667982f40ff63d710fad86), but that makes this Dockerfile very hard to manage since we'd have to account for the whole dependency tree.

`pip install` has a `--platform` option, but it requires either `--only-binary=:all:` or `--no-deps`, none of which makes this useful to us.

This is [a work in progress](https://github.com/nextstrain/docker-base/commit/3aa7a91a4cbd33a47a4bfd040128c20c7b6ffb3b), and I'm still trying to fully grasp these concepts. There's probably a better way to do this.

Dockerfile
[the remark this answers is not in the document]
victorlin: Moving cross-compilation discussion to https://github.com/nextstrain/docker-base/pull/50#discussion_r986331019.

Dockerfile
tsibley: Though it should be possible, it can be tricky to make the whole toolchain work for cross-compilation, esp. across ecosystems like pip/npm/etc.

victorlin: After some extra thought on this, I think the two-builder-stage approach is the best we can get that's still reasonably maintainable. Testing that on a separate branch [victorlin/arm64-cross-compilation](https://github.com/nextstrain/docker-base/tree/victorlin/arm64-cross-compilation), and we can see if the decreased build time (if any) is enough to warrant the extra complexity.

Maybe I'm just bad at searching, but I can't find any examples of other Dockerfiles in the wild that:

1. Use cross-compilation
2. Install a variety of tools via package managers such as `pip` or `npm`

.github/workflows/ci.yml
victorlin: It seems like emulation is supported in the Docker image. This means `amd64` binaries can run on `arm64` without error, though it might be slower. I think emulation is acceptable if the tool cannot easily be built for the target platform, as long as emulation is noted in the Dockerfile.

I'll resolve this comment once each one of these references are addressed by either (1) finding the proper download, (2) trying to build from source, or (3) adding a comment noting on emulation.

tsibley: > It seems like emulation is supported in the Docker image. This means `amd64` binaries can run on `arm64` without error, though it might be slower.

Hmm. This feels artifactual to me. Like that it's a result of having enabled QEMU on the host system for building the image, or something similar. Knowing details about how Docker works, it's not clear to me how it would be generally supported by Docker on all platforms.

Are you testing in CI or on your own M1 system? Is this the macOS Rosetta 2 emulation layer coming into play? How does that work inside the VM that Docker uses behind the scenes to run Linux containers?

victorlin: Good point, I was running this on M1 which does have Rosetta 2 emulation. I wish it were easy to disable and confirm directly, but [it isn't](https://developer.apple.com/forums/thread/669486?answerId=681889022#681889022).

Alternatively, through an `x64` GitHub Codespace, I observed behavior both with and without emulation. When using an `arm64` image, calling executables indeed doesn't work until after emulation is enabled. Note the outputs as comments:

```sh
IMAGE=nextstrain/base:branch-victorlin-add-arm64-keep-scripts
docker pull --platform linux/arm64 $IMAGE

docker run $IMAGE python -c "import platform; print(platform.platform())"
# WARNING: The requested image's platform (linux/arm64) does not match the detected host platform (linux/amd64) and no specific platform was requested
# exec /sbin/entrypoint: exec format error

docker run $IMAGE node -e "console.log(process.arch)"
# WARNING: The requested image's platform (linux/arm64) does not match the detected host platform (linux/amd64) and no specific platform was requested
# exec /sbin/entrypoint: exec format error

# Enable emulation
docker run --rm --privileged tonistiigi/binfmt:latest --install arm64
# installing: arm64 OK

docker run $IMAGE python -c "import platform; print(platform.platform())"
# WARNING: The requested image's platform (linux/arm64) does not match the detected host platform (linux/amd64) and no specific platform was requested
# Linux-5.4.0-1090-azure-aarch64-with-debian-10.13

docker run $IMAGE node -e "console.log(process.arch)"
# WARNING: The requested image's platform (linux/arm64) does not match the detected host platform (linux/amd64) and no specific platform was requested
# arm64
```

So you're right, it's still up to the Docker host machine to support emulation. But I think that's something we can live with (if necessary) if the primary users of the non-`x64` images will be users of Apple silicon Macs, which have access to Rosetta 2.

Dockerfile
victorlin: Done in f60d84d.

devel/build-to-localhost
victorlin: Used `getopts` in [force-push](https://github.com/nextstrain/docker-base/compare/a7a2ac6c0499e4f9adf6d89956329f0461b5dc67..7cec670719ad249f2088f34fc28d05ca3821e612).

Dockerfile
victorlin: Cross-compilation seems to work, and the build time is substantially decreased. I think the changes are complex enough to put in a separate PR: #98

.github/workflows/ci.yml
tsibley: > But I think that's something we can live with (if necessary) if the primary users of the non-`x64` images will be users of Apple silicon Macs, which have access to Rosetta 2.

Agreed. Thanks for digging a bit into this above!

.github/workflows/ci.yml
victorlin: Done: d4f91c4ecf7215a37fda54f19afdf3d08444b7ad

Dockerfile
```diff
@@ -168,6 +168,15 @@ RUN if [ "$TARGETPLATFORM" = "linux/arm64" ]; then \
       ; \
     fi
 
+# Install jaxlib on linux/arm64
+# jaxlib, an evofr dependency, does not have official pre-built binaries for
+# linux/arm64. A GitHub user has provided them in a fork repo.
+# https://github.com/google/jax/issues/7097#issuecomment-1110730040
+RUN if [ "$TARGETPLATFORM" = "linux/arm64" ]; then \
+      pip3 install https://github.com/yoziru/jax/releases/download/jaxlib-v0.3.25/jaxlib-0.3.25-cp310-cp310-manylinux2014_aarch64.manylinux_2_17_aarch64.whl \
```
tsibley, line 176: _non-blocking_

Don't love using a repackaging by a third-party user… I guess the other option is repackaging it ourselves?

.github/workflows/ci.yml
```diff
@@ -38,7 +40,7 @@ jobs:
         username: nextstrain-bot
         password: ${{ secrets.GH_TOKEN_NEXTSTRAIN_BOT_MANAGE_PACKAGES }}
 
-    - run: ./devel/build -p linux/amd64 -r ghcr.io -t $TAG
+    - run: ./devel/build -p linux/amd64,linux/arm64 -r ghcr.io -t $TAG
```
tsibley, line 43: _non-blocking_

Can we build for the two platforms separately and in parallel in different matrix jobs?

tsibley (approved):
\o/ nicely done. I've read thru all the latest commits again and it looks great. One small typo fix and a bunch of non-blocking comments (though I do think some of thing should be addressed).

README.md
```diff
@@ -93,7 +93,8 @@ To add a software program to `nextstrain/base`, follow steps in this order:
 2. Check if it is available via PyPI. You can search on [PyPI's
    website](https://pypi.org/search/). If available, add an install command to
    the section labeled with `Install programs via pip`.
-3. Check if a pre-built binary is available on the software's website (e.g.
+3. Check if a pre-built binary for the `linux/amd64` platform (name contains
+   `linux` and `arm64`/`x86`) is available on the software's website (e.g.
```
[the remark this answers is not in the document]
victorlin: Good catch, done: 86cf790e2084666cf2a5e29a1676251ca917c8b7

Dockerfile
```diff
@@ -53,10 +53,11 @@ RUN mkdir -p /final/bin /final/share /final/libexec
...
 
 # Build RAxML
-# AVX should be widely-supported enough
+# linux/arm64 does not support -mavx and -msse3 compilation flags which are used in the official repository.
+# Make these changes in a fork for now.
+# TODO: Use the official repository if this PR is ever merged: https://github.com/stamatak/standard-RAxML/pull/50
 WORKDIR /build/RAxML
-RUN curl -fsSL https://api.github.com/repos/stamatak/standard-RAxML/tarball/v8.2.12 \
-  | tar xzvpf - --no-same-owner --strip-components=1 \
+RUN git clone --recurse-submodules --single-branch --branch simde https://github.com/nextstrain/standard-RAxML.git . \
```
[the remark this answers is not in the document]
victorlin: Done: 12758b7dfb588976fc36b95c2693179fd11253a9

This is also one less reliance on `git` to be aware of.

builder-scripts/get-nextclade-platform
```diff
@@ -0,0 +1,23 @@
...
+TARGETPLATFORM=$1
+
+case $TARGETPLATFORM in
+    "linux/amd64")
+        echo "x86_64-unknown-linux-gnu"
+        ;;
+    "linux/arm64")
+        echo "aarch64-unknown-linux-gnu"
+        ;;
+esac
```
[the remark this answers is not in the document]
victorlin: Done: f8ed20413a8f520ac1d538ac4c3ff6c347c4b5b2

builder-scripts/get-nextclade-platform
```diff
@@ -0,0 +1,23 @@
...
+# The Docker target platform is provided by the TARGETPLATFORM variable.
+# For more info, see https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope
+#
+# The Nextclade platform should reflect one of the pre-built binaries
+# in the GitHub Release download options, e.g. at the bottom of
+# https://github.com/nextstrain/nextclade/releases/tag/2.5.0
+#
+set -euo pipefail
+
+TARGETPLATFORM=$1
```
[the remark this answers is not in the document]
victorlin: Oh you're right, it should be according to [docs](https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope). Done: f8ed20413a8f520ac1d538ac4c3ff6c347c4b5b2

Dockerfile
```diff
@@ -182,16 +182,12 @@ COPY builder-scripts/ /builder-scripts/
 
 # Download Nextalign v2
 # Set default Nextalign version to 2
-# NOTE: Running this program requires support for emulation on the Docker host
-# if the processor architecture is not amd64.
-RUN curl -fsSL -o /final/bin/nextalign2 https://github.com/nextstrain/nextclade/releases/latest/download/nextalign-x86_64-unknown-linux-gnu \
+RUN curl -fsSL -o /final/bin/nextalign2 https://github.com/nextstrain/nextclade/releases/latest/download/nextalign-$(/builder-scripts/get-nextclade-platform $TARGETPLATFORM) \
```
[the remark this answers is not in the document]
victorlin: Good idea, done: f8ed20413a8f520ac1d538ac4c3ff6c347c4b5b2

Dockerfile
```diff
@@ -147,6 +147,27 @@ RUN curl -fsSL https://github.com/lh3/minimap2/releases/download/v2.24/minimap2-
...
+# First, add system deps for building¹:
+# - libopenblas-dev: Contains optimized versions of BLAS and LAPACK.
+# - libsuitesparse-dev: Contains SuiteSparse.
+#
+# Then, "install" (build) separately since the process requires a special
+# environment variable².
+#
+# ¹ https://cvxopt.org/install/#building-and-installing-from-source
+# ² https://github.com/cvxopt/cvxopt/issues/125#issuecomment-407396491
+RUN if [ "$TARGETPLATFORM" = "linux/arm64" ]; then \
```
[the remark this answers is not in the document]
victorlin: Oh, `bash` is already set to be used: https://github.com/nextstrain/docker-base/blob/4d610805aa58141eb8e71ff34cf14f5f7b9f8f5e/Dockerfile#L11

Updated syntax: f8ed20413a8f520ac1d538ac4c3ff6c347c4b5b2...d85116144bfe90db08390c48d4a1860cf2223689

.github/workflows/ci.yml
victorlin: I believe it's possible, though would require manual pushing/pulling of intermediate images and stringing them together with `docker manifest` to produce a single multi-platform image ("the old way" as described in [this blog post](https://dev.to/cloudx/multi-arch-docker-images-the-easy-way-with-github-actions-4k54)).

This is a trade-off between implementation complexity and build time (which I assume is the main benefit from running in parallel). Currently, the single-platform amd64 build takes [~6 minutes](https://github.com/nextstrain/docker-base/actions/runs/3700380538/jobs/6268749382) while the amd64+arm64 build takes [>1 hour](https://github.com/nextstrain/docker-base/actions/runs/3632781633/jobs/6129028596) on a single job. That means running in parallel would in theory only result in a ~6 minute decrease, which I don't think is enough of an improvement to warrant the increased complexity.

I think BuildKit itself takes advantage of the number of cores on the runner. So another option for parallel platform builds would be to use a runner with more cores.

Dockerfile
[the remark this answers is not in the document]
victorlin: Oops, that didn't work. The tarball endpoint creates a `simde` directory but doesn't provide its contents. I suppose this could work:

```sh
curl -fsSL https://api.github.com/repos/nextstrain/standard-RAxML/tarball/simde \
  | tar xzvpf - --no-same-owner --strip-components=1 \
 && curl -fsSL https://api.github.com/repos/simd-everywhere/simde-no-tests/tarball/294f44c9e948640f9e7d8a2ef7d6d196e14f26eb \
  | tar xzvpf - --no-same-owner --strip-components=1 -C simde \
 && make -f Makefile.AVX.PTHREADS.gcc \
 && cp -p raxmlHPC-PTHREADS-AVX /final/bin
```

but that's a bit too much repo-specific logic to me. Will revert back to `git clone` while taking your suggestions for shallow cloning.

.github/workflows/ci.yml
tsibley: Sounds good. Thanks for explaining!

> …which I don't think is enough of an improvement to warrant the increased complexity.

Agreed!

> So another option for parallel platform builds would be to use a runner with more cores.

Good to keep in mind if (when) the build times get us down.

tsibley (approved):

Dockerfile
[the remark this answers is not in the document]
tsibley: Ahh! I didn't expect/realize our `simde` branch introduced a new submodule, so I figured the submodules implied by `--recurse-submodules` were also bundled up in the previous tarball.

I guess we could vendor it in our `simde` branch instead of submodule-ing it, but also ok with reverting to `git clone` here.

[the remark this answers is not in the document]
victorlin: Actually, I'll make a couple changes here:

1. Add a commit to copy its contents into the `simde` branch, removing the need for submodules
2. Use the tarball endpoint with a commit hash instead of `simde` (and link to the branch in the comment above)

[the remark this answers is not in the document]
victorlin: Done: 6b305c74392c929ba20ba78d3cab21e0101f1ac6...258050a3e1ecfb438221d5f4f9c5f747ab89fb54

Dockerfile
victorlin: Yeah, that could be another adventure so I'll create a separate issue once this PR is merged.

victorlin:
Planning to merge after CI passes and I successfully run through manual testing.

victorlin:
Update: the push-triggered CI failed on the build job with a seemingly transient [network error](https://github.com/nextstrain/docker-base/actions/runs/3707816543/jobs/6284686668). It's been >3 hours and still running. PR-triggered CI took [>4 hours](https://github.com/nextstrain/docker-base/actions/runs/3707816741/jobs/6284689775) on `./devel/build`. Looking at the raw logs, the long build times seem related to #115 (a few notable lines with timestamps pasted below):

```
2022-12-15T21:22:55.0645526Z #104 [linux/arm64 builder 44/47] RUN /builder-scripts/download-repo https://github.com/nextstrain/auspice release .  && npm update && npm install && npm run build && npm link
2022-12-15T21:51:04.8952185Z #104 1689.9 added 1528 packages, and audited 1529 packages in 28m
2022-12-15T21:55:10.3682075Z #104 1935.4 > node auspice.js build --verbose
2022-12-15T22:55:13.8493368Z #104 5538.7 changed 1 package, and audited 1529 packages in 1h
2022-12-15T22:55:39.0958178Z #104 5564.0 > node auspice.js build --verbose
2022-12-16T00:48:18.3439466Z #104 12323.4 added 1 package, and audited 3 packages in 58m
2022-12-16T00:48:18.8022211Z #104 DONE 12323.7s
```

I believe this is because Node.js 14 only came in `amd64` flavors and thus somehow ran "natively" on the build platform, even with the implicit `--platform=$TARGETPLATFORM`. Now with Node.js 16 (which supports `arm64` flavors), it is running with emulation which is much slower.

My opinions:

- 1 hour to build is somewhat acceptable, but 4 hours is not.
- There may be a slight runtime improvement with Node.js 14 → 16, but not enough to justify the drawbacks of increased build time.
- This Node.js upgrade is more fit to come alongside [#98](https://github.com/nextstrain/docker-base/pull/98).

Based on the above points, my planned approach:

1. Revert back to Node.js 14 in this PR for an acceptable build time.
2. Confirm that upgrading to Node.js 16 works in [#98](https://github.com/nextstrain/docker-base/pull/98) without negatively impacting build time.
3. Merge this PR.
4. The Node.js 16 upgrade will happen with merge of [#98](https://github.com/nextstrain/docker-base/pull/98).

tsibley:
@victorlin Sounds like a good plan.  From the log snippet, it seems we're running

```
node auspice.js build --verbose
```

twice? Once as part of `npm install` (since the `prepare` [lifecycle script](https://docs.npmjs.com/cli/v8/using-npm/scripts#life-cycle-scripts) in package.json is `npm run build`) and once explicitly as `npm run build` in our Dockerfile `RUN` line. We can drop the latter then, right?

victorlin:
@tsibley that sounds right! I'll do it in a separate PR.

Dockerfile
victorlin: Moved to #119.

PR #65, opened 2022-07-05
feat: add zstd to runtime installs

### LINKED ISSUE #64: ENH: Add zstd to image
### Context
`zstd` is much faster for decompression than `xz` and compresses just as fast and efficiently.

So there's a case to move to zstd.

### Description
`zstd` seems to be unavailable in our nextstrain base image. Let's add it?

```bash
$ nextstrain:/nextstrain/build $ zstd
bash: zstd: command not found
```

corneliusroemer:
resolves #64

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -143,7 +143,8 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         ruby \
         wget \
         xz-utils \
-        zip unzip
+        zip unzip \
+        zstd
 
 # Configure the prompt for interactive usage
 COPY prompt.sh /etc/profile.d/


corneliusroemer:
Seems to work - but it installs only version 1.3.8 - rather than the most recent 1.5.2

Seems to be because we're using Debian 10 aka buster, which is close to end of life.

Is it normal that you only get 2y old versions of things like zstd on Debian? @tsibley ?

tsibley (approved):
Yes, older releases of software from the package repos are expected on older Debian/Ubuntu releases. We can get a newer zstd version when we upgrade the underlying OS version too, unless there's a particular reason we need a newer version right away.

FWIW for future commits, the rationale you put in #64 is the sort of context/reasoning that's nice to put directly in the commit message itself.

On a broader "moving to zstd" note, it'll also be important to make zstd available in our other runtimes (not just this Docker runtime).

corneliusroemer:
I'll try to move context in the commit message in the future!
I'm not sure how important it is to have the very latest zstd release - it's probably nice to have but not essential.

Could it be that the releases available via apt are those that are able to be built from source without dependency updates? My intuition is that if we just got the binary, we could very well use the latest release.

If we switch to Debian 11 we can go from 1.3 to 1.4, still short of 1.5.2 but closer.

tsibley:
> Could it be that the releases available via apt are those that are able to be built from source without dependency updates?

Ah, no, it's about when (non-security) updates to the non-current Debian/Ubuntu releases stop, which is determined by their package management policies and broader release lifecycle policies.

PR #76, opened 2022-08-03
chore: unpin nextstrain cli

### LINKED ISSUE #52: Nextstrain CLI is pinned at version that's 2 years old
### Current Behavior
https://github.com/nextstrain/docker-base/blob/163cba5590d615d089c0c881077eb5033a9929e0/Dockerfile#L65-L66

### Expected behavior
Nextstrain CLI inside the image is kept up-to-date similar to Augur, Auspice, etc.

At the very least, bump the pin to a more recent version.

corneliusroemer:
resolves #52 (see there for discussion)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -125,7 +125,7 @@ RUN pip3 install envdir==1.0.1
 RUN pip3 install awscli==1.18.195
 
 # Install our own CLI so builds can do things like `nextstrain deploy`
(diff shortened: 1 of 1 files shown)


tsibley (approved):
_non-blocking_

In my ideal change for this, the rationale and auditing in #52 would be included in the commit message here, but it's a relatively minor thing in this case.

PR #78, opened 2022-08-08
Support tag in build script, auto-push `branch-…` tags

### LINKED ISSUE #68: CI: Push `branch-…` tag before test

With the test added in #63, local debugging of a failure requires building the image locally. The local build is unnecessary if the built image is pushed before test failure. `latest` should still be pushed upon successful tests (and other existing conditions).

victorlin:

See commit messages. A brief overview of differences to the CI process:

#### Current CI process

1. Build with default tag `latest`.
2. Test on `latest`.
3. Add named tag (`build-…`/`branch-…`).
4. Push named tag (and `latest` if running on `master`).

#### New CI process

1. Build with named tag (`build-…`/`branch-…`). Push `branch-…` directly.
2. Test on named tag.

If running on `master`:

3. Add `latest` tag.
4. Push `build-…` and `latest` tags.

### Related issue(s)

- Closes #68

### Testing

- [x] CI passes
- [x] Check that `latest` doesn't get pushed in CI runs
- [ ] Post-merge: check that `build-…` and `latest` tags are only pushed after tests pass


devel/tag-latest
```diff
@@ -0,0 +1,26 @@
...
+    exit 1
+fi
+
+tag=$1
+
+BASE_IMAGE="nextstrain/base"
+BASE_BUILDER_IMAGE="nextstrain/base-builder"
+
+docker tag $BASE_BUILDER_IMAGE:{$tag,latest}
+docker tag $BASE_IMAGE:{$tag,latest}
```
victorlin, line 20: [This currently doesn't work](https://github.com/nextstrain/docker-base/runs/7735250010?check_suite_focus=true#step:11:14):

> Error response from daemon: No such image: nextstrain/base-builder:branch-victorlin-auto-push-named-tag

Seems like the tag is ephemeral in [build-tag-push](https://github.com/nextstrain/docker-base/blob/1e27068d571a5138967c21a338c478bff0dbea44/devel/build-tag-push).

.github/workflows/ci.yml
```diff
@@ -24,7 +22,13 @@ jobs:
...
       run: echo "TAG=branch-${GITHUB_REF_NAME//[^A-Za-z0-9._-]/-}" | tee -a $GITHUB_ENV
 
-    - run: ./devel/tag $TAG
+    - if: ${{ github.event_name != 'pull_request' }}
+      uses: docker/login-action@v1
+      with:
+        username: ${{ secrets.DOCKER_LOGIN }}
+        password: ${{ secrets.DOCKER_PASSWORD }}
+
+    - run: ./devel/build-tag-push $TAG
```
victorlin, line 31: This needs more careful thinking. Now, the PR run fails on `build-tag-push` even though it's just the "push" part that won't work (build is fine). Push should be skipped on PR runs.

devel/tag-latest
victorlin: This was due to the switch from `--load` to `--push`. This is no longer an issue after switching back to `--load`.

.github/workflows/ci.yml
victorlin: This is addressed by 3ad8608cd8da0e833c482f3e27f2d77552c976ad + fixups.

.github/workflows/ci.yml
```diff
@@ -22,13 +22,20 @@ jobs:
...
+      run: echo "DIRECT_PUSH=false" | tee -a $GITHUB_ENV
+
     - if: ${{ github.event_name != 'pull_request' }}
       uses: docker/login-action@v1
       with:
         username: ${{ secrets.DOCKER_LOGIN }}
         password: ${{ secrets.DOCKER_PASSWORD }}
 
-    - run: ./devel/build $TAG
+    - run: ./devel/build $TAG $DIRECT_PUSH
```
tsibley, line 38: This logic in the workflow and the additional logic required in `./devel/build` seems to me to be making things more complex than they need to be. The intent of the programs in `./devel/…` is to be composable and easily usable both by CI and local dev. Making them do multiple things is at cross-purposes with that.

Instead of making `./devel/build` optionally push, what about using the existing `./devel/push` instead? e.g. get rid of the `DIRECT_PUSH` var and the changes to `./devel/build` and instead do something like:

```yaml
- if: github.event_name != 'pull_request'
  run: ./devel/push $TAG
```

after the `./devel/build` step here.

devel/build
```diff
@@ -3,12 +3,22 @@
...
 # information.
 #
 set -euo pipefail
 
+# If the tag is not provided, default to "latest".
+if [[ $# -eq 0 ]]; then
+    tag="latest"
+else
+    tag=$1
+fi
```
tsibley, line 21: _non-blocking_

A more concise way of putting this is

```bash
tag="${1:-latest}"
```

tsibley (changes requested):

devel/build
victorlin: Done

.github/workflows/ci.yml
victorlin: Thanks for catching me here, I just reworked it and am happier with the new approach.

Still keeping the option to provide a tag to `./devel/build` since I find the current usage of `latest` to be a bit complex.

.github/workflows/ci.yml
```diff
@@ -24,7 +22,17 @@ jobs:
...
-    - run: ./devel/tag $TAG
+    - run: ./devel/build $TAG
+
+    - if: ${{ github.event_name != 'pull_request' }}
+      uses: docker/login-action@v1
+      with:
+        username: ${{ secrets.DOCKER_LOGIN }}
+        password: ${{ secrets.DOCKER_PASSWORD }}
+
+    - if: ${{ github.event_name != 'pull_request' && github.ref_type == 'branch' && github.ref_name != github.event.repository.default_branch }}
```
tsibley, line 33:
```suggestion
    - if: startsWith(env.TAG, 'branch-')
```

.github/workflows/ci.yml
```diff
@@ -35,15 +43,8 @@ jobs:
         pip install nextstrain-cli
         nextstrain build --image nextstrain/base:$TAG zika-tutorial -F
 
-    - if: ${{ github.event_name != 'pull_request' }}
-      uses: docker/login-action@v1
-      with:
-        username: ${{ secrets.DOCKER_LOGIN }}
-        password: ${{ secrets.DOCKER_PASSWORD }}
-
     - if: ${{ github.event_name != 'pull_request' && github.ref_type == 'branch' && github.ref_name == github.event.repository.default_branch }}
```
tsibley, line 46:
```suggestion
    - if: startsWith(env.TAG, 'build-')
```

tsibley (approved):
Looks good! Two small suggested changes which I think make the behaviour a bit clearer/self-documenting (e.g. of behaviour like "the `latest` tag is the most recent `build-*` tag") and don't repeat the long conditional to determine if this is a release build or not.

(Note that the `${{ … }}` around `if:` keys [is unnecessary](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsif).)

.github/workflows/ci.yml
victorlin: Done: f1f9cc6e27bc5ef8ec52300f14bbb29f653470b3

.github/workflows/ci.yml
victorlin: Done: f1f9cc6e27bc5ef8ec52300f14bbb29f653470b3

.github/workflows/ci.yml
```diff
@@ -24,7 +22,17 @@ jobs:
...
+
+    - if: ${{ github.event_name != 'pull_request' }}
+      uses: docker/login-action@v1
+      with:
+        username: ${{ secrets.DOCKER_LOGIN }}
+        password: ${{ secrets.DOCKER_PASSWORD }}
+
+    - if: startsWith(env.TAG, 'branch-')
+      name: Push $TAG (non-default branch)
+      run: ./devel/push $TAG
```
victorlin, line 35: [CI on PR-triggered run failed](https://github.com/nextstrain/docker-base/runs/7983055498?check_suite_focus=true):

> denied: requested access to the resource is denied

Seems transient, [re-running](https://github.com/nextstrain/docker-base/runs/7983378832?check_suite_focus=true).

victorlin: Oops, of course PR-triggered run failed, it's not supposed to attempt push. Will fix.

victorlin: [Fixed](https://github.com/nextstrain/docker-base/compare/f1f9cc6e27bc5ef8ec52300f14bbb29f653470b3..b97688f85d6bb8ebfce3892b6c0922f551348988)

devel/push
```diff
@@ -5,14 +5,25 @@
...
+BASE_BUILDER_IMAGE="nextstrain/base-builder"
+
 for tag in "$@"; do
-    docker push nextstrain/base-builder:$tag
-    docker push nextstrain/base:$tag
+
+    if [[ $(docker image inspect --format "{{.RepoDigests}}" $BASE_IMAGE:$tag) != '[]' || $(docker image inspect --format "{{.RepoDigests}}" $BASE_BUILDER_IMAGE:$tag) != '[]' ]]; then
+        echo "At least one of the tagged images has already been pushed. This can happen if the newly built image is not available in the local registry." >&2
+        exit 1
+    fi
```
victorlin, line 25: [CI failed post-merge on this error](https://github.com/nextstrain/docker-base/runs/7984133243?check_suite_focus=true). Will investigate.

victorlin: Fixing in #79.

PR #79, opened 2022-08-23
Check all tags before pushing

victorlin:

This check was placed incorrectly in 7d4daa92af7269240a349bca5c56589f27563bee.

Two tags can point to the same image and if one of them is pushed first,
the other will error even though both were "new" at the start of the
script.

### Related issue(s)

- Addresses issue in https://github.com/nextstrain/docker-base/pull/78#discussion_r953171715

### Testing

- [x] CI passes
- [ ] Post-merge: check that `build-…` and `latest` tags are pushed successfully and only after tests pass

### DIFF
--- a/devel/push
+++ b/devel/push
@@ -5,7 +5,7 @@
 # By default this publishes the "latest" tag, but you can provide other tags in
 # addition to or instead of "latest".
 #
-# Errors if a tagged image has already been pushed.
+# Errors if any of the tagged images have already been pushed.
 #
 set -euo pipefail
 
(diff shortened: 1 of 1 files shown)


devel/push
```diff
@@ -19,7 +19,7 @@ BASE_BUILDER_IMAGE="nextstrain/base-builder"
 
 for tag in "$@"; do
 
-    if [[ $(docker image inspect --format "{{.RepoDigests}}" $BASE_IMAGE:$tag) != '[]' || $(docker image inspect --format "{{.RepoDigests}}" $BASE_BUILDER_IMAGE:$tag) != '[]' ]]; then
+    if [[ $tag != 'latest' && ($(docker image inspect --format "{{.RepoDigests}}" $BASE_IMAGE:$tag) != '[]' || $(docker image inspect --format "{{.RepoDigests}}" $BASE_BUILDER_IMAGE:$tag) != '[]') ]]; then
```
victorlin, line 22: Interestingly, [in the failing run](https://github.com/nextstrain/docker-base/runs/7984133243?check_suite_focus=true) it looks like the `latest` pushes succeeded and it's actually the `build-` tag that exited before push. So I'm not sure that this is the right change. Will need to do some local debugging to verify.

tsibley: I don't think this is the right change, though it may fix the symptom in our usage of `./devel/push`. What's happening, I believe, is that `./devel/push latest build-…` gets run, this check passes for `latest` tag, the images tagged `latest` are pushed, and then on the next iteration of this `for` loop this check fails for the `build-…` tag because it refers to the same `latest` image which was just pushed.

One possible fix, I think, would be to loop over all tags and make this assertion. Then loop again to push.

tsibley (changes requested):

devel/push
victorlin: I just checked locally and you're right, `RepoDigests` has a value if that local image has already been pushed regardless of tag. I'll update with your proposed fix.

tsibley (approved):

PR #83, opened 2022-09-12
Run as non-root "nextstrain" user and set HOME=/nextstrain

tsibley:
See commit messages.

### Testing
- [x] CI passes
- [x] Manually test real (i.e. complex) builds aren't affected:
  - [x] ncov
  - [x] monkeypox
  - [x] zika

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -245,6 +245,10 @@ COPY --from=builder /nextstrain /nextstrain
 COPY entrypoint entrypoint-aws-batch /sbin/
 RUN chmod a+rx /sbin/entrypoint*
 
+# Make /nextstrain a global HOME, writable by any UID (like /tmp)
+RUN chmod a+rwXt /nextstrain
+ENV HOME=/nextstrain
+
 # Setup a non-root user for normal operations
 RUN useradd nextstrain \
     --system \
--- a/Dockerfile
+++ b/Dockerfile
@@ -245,6 +245,15 @@ COPY --from=builder /nextstrain /nextstrain
 COPY entrypoint entrypoint-aws-batch /sbin/
 RUN chmod a+rx /sbin/entrypoint*
 
+# Setup a non-root user for normal operations
+RUN useradd nextstrain \
+    --system \
+    --user-group \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -245,6 +245,19 @@ COPY --from=builder /nextstrain /nextstrain
...
+ENV HOME=/nextstrain
+
+# Setup a non-root user for normal operations
+RUN useradd nextstrain \
+    --system \
+    --user-group \
+    --shell /bin/bash \
+    --home-dir /nextstrain \
+    --no-log-init
+USER nextstrain:nextstrain
```
victorlin, line 259: Noting that I don't see a name on the user in `nextstrain shell`, but this was already the case.

```
$ touch ~/test

$ ls -l ~
total 12
drwxr-xr-x 11 root root    4096 Sep 12 21:34 augur
drwxr-xr-x 14 root root    4096 Sep 12 21:38 auspice
drwxr-xr-x  8  502 dialout  256 Sep 16 20:59 build
drwxr-xr-x 11 root root    4096 Sep 12 21:34 fauna
-rw-r--r--  1  502 dialout    0 Sep 16 21:05 test

$ id
uid=502 gid=20(dialout) groups=20(dialout)
```

victorlin (approved):
Home directory change checks out:

```sh
nextstrain shell --image nextstrain/base:branch-trs-non-root-user .
cd ~
pwd
# /nextstrain
```

Dockerfile
tsibley: Yeah, that's because the Docker runner launches the container with your current uid/gid from the host kernel (so 502:20 in the above, which happen to coincide with no user and the `dialout` group in the container's `/etc/passwd` and `/etc/group`).

(The name in the container is really only because a name is necessary.)

tsibley: and to clarify, 502:20 are probably the uid:gid in the Linux helper VM used to run Docker on macOS, not your macOS uid:gid.

victorlin: Got it!

PR #84, opened 2022-09-12
Configure the prompt using bashrc instead of a profile.d snippet

tsibley:
The upshot of this is that any interactive Bash session gets the prompt,
not just a login session.  It also makes overriding the file in newer
`nextstrain shell` versions a bit less custom, since /etc/bash.bashrc is
the standard Debian/Ubuntu system bashrc path.

### Testing
- [x] Manual testing with `nextstrain shell` and `docker run …`
- [x] CI passes

### DIFF
--- a/bashrc
+++ b/bashrc
@@ -1,0 +1,8 @@
+# Note that this file is overridden by newer versions of `nextstrain shell` and
+# so it provides only a fallback prompt for users of older CLI versions or
+# direct image users (e.g. `docker run …`).
+reset="\[\e[0m\]"
+bold="\[\e[1m\]"
+magenta="\[\e[35m\]"
+PS1="${bold}nextstrain:${magenta}\w${reset} ${bold}\$ ${reset}"
+unset reset bold magenta
--- a/Dockerfile
+++ b/Dockerfile
@@ -148,8 +148,8 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         zip unzip \
         zstd
 
-# Configure the prompt for interactive usage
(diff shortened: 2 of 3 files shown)


bashrc
```diff
@@ -0,0 +1,8 @@
+# Note that this file is overridden by newer versions of `nextstrain shell` and
+# so it provides only a fallback prompt for users of older CLI versions or
+# direct image users (e.g. `docker run …`).
```
tsibley, line 3: This comment is a bit forward-looking to the WIP `nextstrain shell` improvements re: the new managed Conda runtime.

bashrc
```diff
@@ -0,0 +1,8 @@
+# Note that this file is overridden by newer versions of `nextstrain shell` and
```
victorlin, line 1: _(non-blocking)_

My text editor (VSCode) doesn't understand that this is a bash script. It tries to be "smart" and thinks it's Markdown instead, likely based on the `#` in the first few lines.

A proper shebang fixes that, though it might not be necessary/appropriate for this kind of file?

victorlin (approved):

bashrc
tsibley: Ah, yeah I didn't include a shebang because rc files are meant to be `source`d not directly invoked. For that same reason, they're not marked executable.

I generally think that file detection stuff should live outside of the files themselves, except in rare circumstances. That is, I'm not a huge fan of using modelines, though some people use them everywhere (and there is [a VS Code plugin](https://marketplace.visualstudio.com/items?itemName=chrislajoie.vscode-modelines) to respect them).

PR #85, opened 2022-09-21
Non-root user, deux

tsibley:
See commit messages for details.

### Related issue(s)
Related to #83 

### Testing
- [x] Manually built image and tested a variety of invocations.
- [x] Test ncov-ingest image build against it
- [x] Test pathogen build against it

### DIFF
--- a/entrypoint
+++ b/entrypoint
@@ -17,6 +17,11 @@ if [[ $# -eq 0 ]]; then
         the source of the \`nextstrain\` command for examples.
     "
 else
-    # Otherwise, exec into whatever command is provided.
-    exec "$@"
+    # Otherwise, exec into whatever command is provided…
+    if [[ "$(id -u):$(id -g)" == 0:0 ]]; then
+        # …switching to nextstrain:nextstrain first if we're root.
+        exec setpriv --reuid nextstrain --regid nextstrain --init-groups "$@"
+    else
+        exec "$@"
+    fi
 fi
--- a/Dockerfile
+++ b/Dockerfile
@@ -249,14 +249,13 @@ RUN chmod a+rx /sbin/entrypoint*
 RUN chmod a+rwXt /nextstrain
 ENV HOME=/nextstrain
 
-# Setup a non-root user for normal operations
+# Setup a non-root user for optional use
 RUN useradd nextstrain \
     --system \
     --user-group \
     --shell /bin/bash \
     --home-dir /nextstrain \
     --no-log-init
-USER nextstrain:nextstrain
 
 # The host should bind mount the pathogen build dir into /nextstrain/build.
 WORKDIR /nextstrain/build
--- a/Dockerfile
+++ b/Dockerfile
@@ -143,6 +143,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         nodejs \
         perl \
         ruby \
+        util-linux \
         wget \
         xz-utils \
         zip unzip \


entrypoint
```diff
@@ -17,6 +17,11 @@ if [[ $# -eq 0 ]]; then
         the source of the \`nextstrain\` command for examples.
     "
 else
-    # Otherwise, exec into whatever command is provided.
-    exec "$@"
+    # Otherwise, exec into whatever command is provided…
+    if [[ "$(id -u):$(id -g)" == 0:0 ]]; then
```
joverlee521, line 21: [TIL](https://superuser.com/questions/626843/does-the-root-account-always-have-uid-gid-0) "root" is just a name but uid will always be 0.

entrypoint
```diff
@@ -17,6 +17,11 @@ if [[ $# -eq 0 ]]; then
...
+        # …switching to nextstrain:nextstrain first if we're root.
+        exec setpriv --reuid nextstrain --regid nextstrain --init-groups "$@"
```
joverlee521, line 23: I have no idea what `setpriv` does so I went to read the [docs](https://man7.org/linux/man-pages/man1/setpriv.1.html) and saw this bit:

> Setting a uid or gid does not change capabilities, although
           the exec call at the end might change capabilities. This
           means that, if you are root, you probably want to do
           something like:
>
> setpriv --reuid=1000 --regid=1000 --inh-caps=-all

Do this apply here?

joverlee521 (approved):
Thanks for making these changes! I still get lost in the Docker world 😅

entrypoint
tsibley: Yep, similar to how PID 1 is special.

entrypoint
tsibley: I'm familiar with the basics of capabilities and a little fuzzier on their specifics, but that recommendation doesn't apply here AFAIK. The Docker daemon which launches the container process is already limiting down capabilities at start, and I think it's fine to keep that limited set inheritable. Note that the former `USER` line in the Dockerfile doesn't do anything related to capabilities. The `setpriv` invocation here is basically ~equivalent to `USER`.

joverlee521:
I [reverted](https://github.com/nextstrain/docker-base/commit/5e89e3cdc768cc27c2c1cf3ddf89f42c9e9f200e) this PR merge since it was causing Permission denied errors for our pipelines on AWS Batch (see [Slack thread](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1663955018231249))

tsibley:
Noting for posterity: Following the revert of the PR above, a0ddb0e was then re-applied as 69fa7296ddbf6cea312119d0c07ca8c4d54ca176 to once again remove the `USER` line. The net result is as if just the first commit of this PR was merged (or alternatively that just the second commit of this PR was reverted). This put the image in a state where it runs as root, e.g. similar to its state prior to b0612f7.

tsibley:
With the latest image using `setpriv` (without `USER`), `nextstrain/base:build-20220922T213434Z`, I was able to reproduce this with a more minimal test case than our production builds:

```
$ md /tmp/empty
$ touch /tmp/empty/foo
$ nextstrain build --aws-batch --cpus 1 --memory 2gb --image nextstrain/base:build-20220922T213434Z --exec bash /tmp/empty -c 'id; ls -ld'
Nextstrain Run ID: 138053cd-f715-4d9b-9fec-3db2914f16ad
Uploading /tmp/empty to S3
zipping: /tmp/empty
zipping: /tmp/empty/foo
uploaded: s3://nextstrain-jobs/138053cd-f715-4d9b-9fec-3db2914f16ad.zip
Submitting job
AWS Batch Job ID: 4e4a532c-304c-4307-9da8-780dbc791af1
Watching job status

Press Control-C twice within 10 seconds to cancel this job,
      Control-Z to detach from it.

Job now SUBMITTED
Job now RUNNING
[batch] [2022-09-23T14:12:00-07:00] + case "$NEXTSTRAIN_AWS_BATCH_WORKDIR_URL" in
[batch] [2022-09-23T14:12:00-07:00] + aws s3 cp --no-progress s3://nextstrain-jobs/138053cd-f715-4d9b-9fec-3db2914f16ad.zip /nextstrain/build.zip
[batch] [2022-09-23T14:12:01-07:00] download: s3://nextstrain-jobs/138053cd-f715-4d9b-9fec-3db2914f16ad.zip to ../build.zip
[batch] [2022-09-23T14:12:01-07:00] + unzip /nextstrain/build.zip
[batch] [2022-09-23T14:12:01-07:00] Archive:  /nextstrain/build.zip
[batch] [2022-09-23T14:12:01-07:00] error:  cannot create foo
[batch] [2022-09-23T14:12:01-07:00]         Permission denied
Job FAILED after 0.1 minutes (exited 50)
Downloading all files modified by job to /tmp/empty
```

The job succeeds if there are no build files to extract (e.g. re-run after `rm /tmp/empty/foo`), showing that while the uid/gid is as expected, `/nextstrain/build` has permissions I forgot to account for after removing the `USER` line:

```
uid=999(nextstrain) gid=999(nextstrain) groups=999(nextstrain)
drwxr-xr-x 2 root root 6 Sep 22 21:44 .
```

tsibley:
The permissions of `/nextstrain/build` don't matter in the local Docker runtime because we mount over the container's copy with the build directory from the host and run as the host's local uid/gid.

But on AWS Batch we don't mount over `/nextstrain/build`. Instead, extract a build dir archive into it at job start. Since the container job runs as the default user, which is root, when our entrypoint switches to nextstrain it's no longer able to write to `/nextstrain/build`.

I think we can keep 8ccca2e with the addition of a couple well-placed `chown`s.

tsibley:
[Third time's the charm?](https://github.com/nextstrain/docker-base/pull/86) :expressionless:

PR #86, opened 2022-09-23
Non-root user, trois

tsibley:
Switch to non-root user (nextstrain) in entrypoint exec chain

Restores default use of a non-root user even for ad-hoc image uses without getting in the way (like USER does) of downstream images inheriting from us (see previous commit).

setpriv is part of util-linux, which is installed by default but is also installed explicitly in case that changes.  It is an exec-chaining program explicitly designed for modifying privs normally retained across an exec boundary.

A chown ensures that /nextstrain/build is writable by the nextstrain user.  This doesn't matter for the local Docker runtime (e.g. via `nextstrain build --docker`) because we mount over the container's /nextstrain/build with the build directory from the host (and run as the host user's uid/gid).  But for the AWS Batch runtime we _don't_ mount over /nextstrain/build: the entrypoint extracts a build archive into it at job start.  Since this happens _after_ the setpriv, /nextstrain/build needs to be writable by the nextstrain user.

### Related issue(s)
Related to #83, #85

### Testing
- [x] CI
- [x] Manual testing locally
- [x] Test ncov-ingest image build against it
- [x] Test local pathogen build against it
- [x] Test AWS Batch pathogen build against it

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -143,6 +143,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         nodejs \
         perl \
         ruby \
+        util-linux \
         wget \
         xz-utils \
         zip unzip \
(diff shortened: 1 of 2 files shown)


tsibley:
I'll plan to see this through next week.

victorlin (approved):

PR #87, opened 2022-09-28
Remove unneeded pull script

victorlin:

When running devel/build, if an image doesn't exist locally, `--cache-from` will pull it from the registry automatically.

### Related issue(s)

Idea conceived in https://github.com/nextstrain/docker-base/pull/50#discussion_r982711895

### Testing

- [x] Checks pass


tsibley (approved):

devel/pull
```diff
@@ -1,18 +0,0 @@
-#!/bin/bash
-#
-# Pull the nextstrain/base and nextstrain/base-builder images from Docker Hub.
```
victorlin, line 3: I just noticed that `--cache-from` only pulls the cache manifest, which is much faster than pulling the image itself.

This new realization wouldn't make a difference for the PR if `./devel/pull` was only intended for use right before `./devel/build`. However, I can see how one might want to use `./devel/pull` as a convenience script to pull both images for a specific tag (e.g. `./devel/pull branch-victorlin-remove-pull`) without writing the separate commands.

I'm thinking of restoring the script itself, but still removing the existing references from CI + README. @tsibley does that sound ok to you?

tsibley: Ah, sure. I think I'm also ok with removing it still if it's solely convenience.

victorlin: I'll just remove it then. I don't see myself using it for now, and it's easy to add back later.

PR #88, opened 2022-09-29
Improve `--cache-from` caching

### LINKED ISSUE #57: Use better caching options to reduce build times
> Switching to buildx should also allow us to use better caching options to reduce build times

_Originally posted by @tsibley in https://github.com/nextstrain/docker-base/issues/49#issuecomment-1159270328_

- [x] Add `--cache-from` for the `branch-*` tag when applicable: ([done](https://github.com/nextstrain/docker-base/pull/88))
- [x] Allow more layers to use caching by placing them before `ARG CACHE_DATE` done: f7fbd3ea75077479d82a8367dd80599ae772d716, a7159ec92fdec26b208e86005004889a511231be

victorlin:

Import caches from both `latest` and `$tag` (if available) on registry for caching.

Reference: https://docs.docker.com/engine/reference/commandline/buildx_build/#cache-from

### Related issue(s)

- Closes #57

### Testing

- [x] Checks pass

### DIFF
--- a/devel/build
+++ b/devel/build
@@ -47,8 +47,8 @@ docker buildx build \
     --platform $platform \
     --build-arg CACHE_DATE \
     --build-arg GIT_REVISION \
-    --cache-from $BASE_BUILDER_IMAGE \
-    --cache-from $BASE_IMAGE \
+    --cache-from $BASE_BUILDER_IMAGE:latest \
+    --cache-from $BASE_IMAGE:latest \
     --cache-to type=inline \
     --tag $BASE_BUILDER_IMAGE:$tag \
     --load \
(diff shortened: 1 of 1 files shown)


devel/build
```diff
@@ -48,7 +48,9 @@ docker buildx build \
     --build-arg CACHE_DATE \
     --build-arg GIT_REVISION \
     --cache-from $BASE_BUILDER_IMAGE:latest \
+    --cache-from $BASE_BUILDER_IMAGE:$tag \
     --cache-from $BASE_IMAGE:latest \
+    --cache-from $BASE_IMAGE:$tag \
     --cache-to type=inline \
```
tsibley, line 54: Is this at all problematic when `$tag == latest`?

tsibley (approved):

devel/build
victorlin: No, I just ran `./devel/build` locally and got this without error:

```
 => importing cache manifest from nextstrain/base-builder:latest                                                                                                                                                      1.0s
 => importing cache manifest from nextstrain/base:latest   
```

PR #91, opened 2022-10-07
Name the final stage explicitly, move scripts, other minor refactoring

victorlin:

See individual commits.

### Related issue(s)

_N/A_

### Testing

- [x] Checks pass


devel/build
```diff
@@ -36,14 +36,14 @@ if ! docker buildx inspect "$builder" &>/dev/null; then
     # However, if this is changed and it was previously run on your machine,
     # you may need to remove the builder manually before running the script:
     #     docker builder rm "nextstrain-builder"
-    docker buildx create --name "$builder" --driver docker-container
+    docker buildx create --name $builder --driver docker-container
```
tsibley, line 39: This is an odd change to be making to me, as quoting variable substitutions is a good defensive practice against future meta-chars in the variable value. I'd expect to be changing the `devel/*` programs to _add_ quotes to variable substitutions!

(There's also still a `"$builder"` form a few lines above in this file, so not sure why that wouldn't be changed here too…)

.github/workflows/ci.yml
```diff
@@ -20,7 +20,7 @@ jobs:
       # and hyphens.
       run: echo "TAG=branch-${GITHUB_REF_NAME//[^A-Za-z0-9._-]/-}" | tee -a $GITHUB_ENV
 
-    - run: ./devel/build $TAG
+    - run: ./docker/build $TAG
```
tsibley, line 23: Hmm. Having a directory called `docker` inside the `docker-base` repo seems a bit arbitrary to me. This whole repo is about Docker, so why are some things in an additional `docker` directory?

What about opting to leave the **devel**opment programs (`push`, `pull`, etc) in `devel/` and instead moving the builder programs (`download-repo`, `latest-augur-release-tag`) into a new `builder/` directory?

devel/build
victorlin: Good point, updated to use quotes: 257851bb4de28b899783b0b2dbe8d6c496a82a52

.github/workflows/ci.yml
victorlin: Updated: 0770e9e048e18273d2ad0d08740ff3519c2cb8c1

devel/build
tsibley: Ok… quotes around all strings feels unnecessary, but it's also fine.

devel/build
```diff
@@ -18,53 +18,54 @@ tag="${1:-latest}"
 export GIT_REVISION=$(git describe --tags --abbrev=40 --always --dirty || true)
 
 # The nextstrain/base Dockerfile is a multi-stage with both a "builder" target
-# and a main target.  To enable proper caching via --cache-from we need both
+# and a main target.  To enable proper caching via --cache-from "we need bo"th
 # these images available to pull layers from.  This means pulling both in at
```
tsibley, line 22: something went wrong here

devel/build
tsibley: I'd also be inclined to make such style changes (esp. those "for consistency" :-) across all files at once instead of just one.

tsibley (approved):
My review comments here are minor, so merge when you feel this is ready.

devel/build
victorlin: Oops, fixed.

devel/build
victorlin: Toned back the quoting and made changes in other files: aa28baf42fb3e0255ab5f2ecad6823e2864cb628

victorlin:
Will merge once CI passes.

PR #92, opened 2022-10-11
Re-organize Dockerfile

victorlin:

Some improvements to the Dockerfile.

### Related issue(s)

- #57
    - Improves caching by handling version-pinned software before `CACHE_DATE`

### Testing

- [x] Checks pass
- [x] Manually check `nextstrain/base:branch-victorlin-organize-dockerfile`:

```sh
docker run nextstrain/base:build-20221013T234036Z ls -lR --time-style='+' /usr/local/{bin,libexec,share} > old.txt
docker run nextstrain/base:branch-victorlin-organize-dockerfile ls -lR --time-style='+' /usr/local/{bin,libexec,share} > new.txt
diff old.txt new.txt
# no diff!
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -110,10 +110,6 @@ RUN pip3 install --editable "/nextstrain/augur"
 RUN pip3 install epiweeks==2.1.2
 
 # Add Pangolin and PangoLEARN + deps
-RUN curl -fsSL https://github.com/virus-evolution/gofasta/releases/download/v0.0.6/gofasta-linux-amd64 \
-        -o /usr/local/bin/gofasta
-RUN cd /usr/local/bin && curl -fsSL https://github.com/lh3/minimap2/releases/download/v2.24/minimap2-2.24_x64-linux.tar.bz2 \
-  | tar xjvpf - --no-same-owner --strip-components=1 minimap2-2.24_x64-linux/minimap2
 RUN pip3 install pysam
 RUN pip3 install git+https://github.com/cov-lineages/pangolin.git@v3.1.17
 RUN pip3 install git+https://github.com/cov-lineages/pangoLEARN.git@2021-12-06
@@ -202,6 +198,14 @@ RUN curl -L https://github.com/shenwei356/csvtk/releases/download/v0.24.0/csvtk_
 # Add seqkit
 RUN curl -L https://github.com/shenwei356/seqkit/releases/download/v2.2.0/seqkit_linux_amd64.tar.gz | tar xz --no-same-owner -C /usr/local/bin
 
+# Add gofasta (for ncov/Pangolin)
+RUN curl -fsSL https://github.com/virus-evolution/gofasta/releases/download/v0.0.6/gofasta-linux-amd64 \
+  -o /usr/local/bin/gofasta
+
+# Add minimap2 (for ncov/Pangolin)
+RUN curl -fsSL https://github.com/lh3/minimap2/releases/download/v2.24/minimap2-2.24_x64-linux.tar.bz2 \
+  | tar xjvpf - --no-same-owner --strip-components=1 -C /usr/local/bin minimap2-2.24_x64-linux/minimap2
+
 # Ensure all container users can execute these programs
 RUN chmod a+rx /usr/local/bin/* /usr/local/libexec/*
 
@@ -221,8 +225,6 @@ COPY --from=builder \
     /usr/local/bin/augur \
     /usr/local/bin/aws \
     /usr/local/bin/envdir \
-    /usr/local/bin/gofasta \
-    /usr/local/bin/minimap2 \
     /usr/local/bin/nextstrain \
     /usr/local/bin/pangolin \
     /usr/local/bin/pangolearn.smk \
--- a/Dockerfile
+++ b/Dockerfile
@@ -111,8 +111,7 @@ RUN pip3 install epiweeks==2.1.2
 
 # Add Pangolin and PangoLEARN + deps
 RUN curl -fsSL https://github.com/virus-evolution/gofasta/releases/download/v0.0.6/gofasta-linux-amd64 \
-        -o /usr/local/bin/gofasta \
-  && chmod a+rx /usr/local/bin/gofasta
+        -o /usr/local/bin/gofasta
 RUN cd /usr/local/bin && curl -fsSL https://github.com/lh3/minimap2/releases/download/v2.24/minimap2-2.24_x64-linux.tar.bz2 \
   | tar xjvpf - --no-same-owner --strip-components=1 minimap2-2.24_x64-linux/minimap2
 RUN pip3 install pysam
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -206,6 +201,14 @@ RUN curl -L https://github.com/shenwei356/csvtk/releases/download/v0.24.0/csvtk_
 # Add seqkit
 RUN curl -L https://github.com/shenwei356/seqkit/releases/download/v2.2.0/seqkit_linux_amd64.tar.gz | tar xz --no-same-owner -C /usr/local/bin
 
+# Add gofasta (for ncov/Pangolin)
+RUN curl -fsSL https://github.com/virus-evolution/gofasta/releases/download/v0.0.6/gofasta-linux-amd64 \
+  -o /usr/local/bin/gofasta
+
```
tsibley: Do we not need to retain the `chmod +x`?

Dockerfile
```diff
@@ -28,41 +28,87 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
 # functionality must live in this Dockerfile. The following dependencies may be
 # used by multiple pathogen workflows, but they have been commented according to
 # the original pathogen that added these dependencies.
 
-# mafft
-WORKDIR /build/mafft
-RUN curl -fsSL https://mafft.cbrc.jp/alignment/software/mafft-7.475-linux.tgz \
-  | tar xzvpf - --no-same-owner --strip-components=2 mafft-linux64/mafftdir/
+# Create directories to be copied in final stage.
+RUN mkdir -p /out/bin /out/share /out/libexec
```
tsibley, line 40: Consider naming this `/final` instead of `/out` to mark that it's for the `final` stage?

Dockerfile
```diff
@@ -28,41 +28,87 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
+# Build vcftools
 WORKDIR /build/vcftools
 RUN curl -fsSL https://github.com/vcftools/vcftools/releases/download/v0.1.16/vcftools-0.1.16.tar.gz \
-  | tar xzvpf - --no-same-owner --strip-components=2
-RUN ./configure --prefix=$PWD/built && make && make install
+  | tar xzvpf - --no-same-owner --strip-components=2 \
+ && ./configure --prefix=$PWD/built \
+ && make && make install \
+ && cp -rp built/bin/*    /out/bin \
+ && cp -rp built/share/*  /out/share
```
tsibley, line 67: Maybe this?

```suggestion
 && ./configure --prefix=/out \
 && make && make install
```

or maybe explicit is better.

Dockerfile
```diff
@@ -79,22 +125,36 @@ RUN pip3 install snakemake==5.10.0
...
-# Add Nextstrain components
+
+# 4. Add Nextstrain components
 
 # Allow caching to be avoided from here on out by calling
 # docker build --build-arg CACHE_DATE="$(date)"
 ARG CACHE_DATE
 
+# Nextclade/Nextalign are downloaded directly but using the latest version, so
+# they belong after CACHE_DATE.
```
tsibley, line 146: I was confused at first because I had seen Nextclade/Nextalign downloaded above, but then realized those are v1 and this is v2. This small change would have helped, I think:

```suggestion
# Nextclade/Nextalign v2 are downloaded directly but using the latest version,
# so they belong after CACHE_DATE (unlike Nextclade/Nextalign v1).
```

Dockerfile
```diff
@@ -79,22 +125,36 @@ RUN pip3 install snakemake==5.10.0
...
+
+# Download Nextalign v2
+RUN curl -fsSL -o /out/bin/nextalign2 https://github.com/nextstrain/nextclade/releases/latest/download/nextalign-x86_64-unknown-linux-gnu
+# Set default Nextalign version to 2
+RUN ln -sv nextalign2 /out/bin/nextalign
+
+# Download Nextclade v2
+RUN curl -fsSL -o /out/bin/nextclade2 https://github.com/nextstrain/nextclade/releases/latest/download/nextclade-x86_64-unknown-linux-gnu
+# Set default Nextclade version to 2
+RUN ln -sv nextclade2 /out/bin/nextclade
```
tsibley, line 156: Why not bundle the `ln` calls here into the same `RUN`?

Dockerfile
```diff
@@ -223,7 +223,7 @@ COPY --from=builder /out/libexec/ /usr/local/libexec/
 ENV MAFFT_BINARIES=/usr/local/libexec
 
 # Ensure all container users can execute these programs
-RUN chmod a+rX /usr/local/bin/* /usr/local/libexec/*
+RUN chmod a+rx /usr/local/bin/* /usr/local/libexec/*
```
tsibley, line 226: The original use of `+X` was intentional as sometimes there are non-executable things in libexec (and more rarely bin) and so I wanted at least one executable bit set to indicate something was supposed to be executable. But it doesn't appear to be the case here, so `+x` is fine.

I do think this is more properly part of the prior commit introducing `/out`, since that change won't work without this one.

tsibley (approved):
Overall, love to see this get some attention. Really like 2ab774c501ff676f19e1d3d33168fc58ee6e33cc in particular. A few notes below, nothing major.

Dockerfile
tsibley: Oh I see. Yes, it needs retaining, but it's omission wasn't noticed because the the change wasn't tested in isolation but with 472f556.

tsibley:
I extended the diff a bit to check the rest of the dirs.

```
docker run nextstrain/base:build-20221005T235113Z ls -lR /usr/local/{bin,libexec,share} > old.txt
docker run nextstrain/base:branch-victorlin-organize-dockerfile ls -lR /usr/local/{bin,libexec,share} > new.txt
```

[Looks fine](https://tsibley.net/paste/2022-10-1394ln8J4e).

Dockerfile
victorlin: This alternative hadn't crossed my mind, thanks for pointing it out. I do think explicit is better here, since otherwise it's hard to tell how `/final/share` is used. Will leave as-is.

Dockerfile
victorlin: Consolidated the `chmod` changes into 48e4ca2a895de6c54b4e7fbcf5321c28ddaf4074.

Dockerfile
victorlin: I like that. Done: cf4fdce6a283f56665e256b46fb8e18e8c135a68

Dockerfile
victorlin: Done: c528ef5cb37a7459ddbfb36efc5cee87eba8757c

Dockerfile
victorlin: Done: 8c3c27c9f5c6dc56863d040c29d60e8411dc4a17

Dockerfile
victorlin: Consolidated the `chmod` changes into 48e4ca2a895de6c54b4e7fbcf5321c28ddaf4074 and moved that before the other changes.

victorlin:
Will merge once CI and the manual diff check passes.

victorlin:
Updated diff check to ignore timestamps and use newer image:

```sh
docker run nextstrain/base:build-20221013T234036Z ls -lR --time-style='+' /usr/local/{bin,libexec,share} > old.txt
docker run nextstrain/base:branch-victorlin-organize-dockerfile ls -lR --time-style='+' /usr/local/{bin,libexec,share} > new.txt
diff old.txt new.txt
# no diff!
```

PR #93, opened 2022-10-13
Annotate package installs

victorlin:

Currently, the reasoning on why each package is used

1. can be deduced from direct references in the Dockerfile (e.g. curl), or
2. can be deduced from git blame (e.g. jq: c256a30eccf08ba5c245656a2b2bc3b1f2f713f9), or
3. is unclear.

To make this easier, document why each package is installed.

### Related issue(s)

_N/A_

### Testing

- [x] Checks pass

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,6 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         curl \
         git \
         jq \
-        libgmp-dev \
         pkg-config \
         zlib1g-dev
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,6 +11,15 @@ FROM python:3.7-slim-buster AS builder
 SHELL ["/bin/bash", "-c"]
 
 # Add system deps for building
+# autoconf, automake: for building VCFtools; may be used by package managers to build from source
+# build-essential: contains gcc, g++, make, etc. for building various tools; may be used by package managers to build from source
+# ca-certificates: for secure HTTPS connections
+# curl: for downloading source files
+# git: for git pip installs
+# jq: used in builder-scripts/latest-augur-release-tag
+# pkg-config: for building VCFtools; may be used by package managers to build from source
+# zlib1g-dev: for building VCFtools; may be used by package managers to build from source
+# nodejs: for installing Auspice
 RUN apt-get update && apt-get install -y --no-install-recommends \
         autoconf \
         automake \
@@ -191,6 +200,15 @@ RUN cd /nextstrain/auspice && npm update && npm install && npm run build && npm 
 FROM python:3.7-slim-buster AS final
 
 # Add system runtime deps
+# bzip2, gzip, xz-utils, zip, unzip, zstd: install compression tools
+# ca-certificates: [Dockerfile] for secure HTTPS connections; may be used by workflows
+# curl: [Dockerfile] for downloading binaries directly; may be used by workflows
+# jq: may be used by workflows
+# less: for usability in an interactive prompt
+# perl: for running VCFtools
+# ruby: may be used by workflows
+# wget: may be used by workflows
+# nodejs: for running Auspice
 RUN apt-get update && apt-get install -y --no-install-recommends \
         bzip2 \
         ca-certificates \
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,8 +20,8 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         git \
         jq \
         libgmp-dev \
-        libpng-dev \
-        pkg-config
+        pkg-config \
+        zlib1g-dev
 
 # Install a specific Node.js version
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -136,6 +146,15 @@ RUN cd /nextstrain/auspice && npm update && npm install && npm run build && npm
 FROM python:3.7-slim-buster
 
 # Add system runtime deps
+# bzip2, gzip, xz-utils, zip, unzip, zstd: install compression tools
+# ca-certificates: ?
+# curl: [Dockerfile] for downloading binaries directly; may be used by workflows
+# jq: ?
```
victorlin, line 152: @corneliusroemer do you remember why you added `jq` in 8a6a29d906de9546c8b2c6be118bb95b206ba868? Is it just for usability in workflows?

Dockerfile
```diff
@@ -11,6 +11,16 @@ FROM python:3.7-slim-buster AS builder
...
 # Add system deps for building
+# autoconf, automake: for building VCFtools; may be used by package managers to build from source
+# build-essential: contains gcc, g++, make, etc. for building various tools; may be used by package managers to build from source
+# ca-certificates: ?
+# curl: for downloading source files
+# git: for git pip installs
+# jq: used in devel/latest-augur-release-tag
+# libgmp-dev: ?
+# libpng-dev: ?
+# pkg-config: ?
```
victorlin, line 22: @tsibley do you remember why the packages with `?` were added in 3917c1b9aa0aa19e694f94f6bd3e56418f248269?

victorlin (commented):
I've filled out what I could find through a quick search. There's still some missing things which I've tagged others to help on.

Dockerfile
tsibley: These would have been for compiling deps. Not sure which ones, possibly ones a) that have since been removed or b) which we no longer compile ourselves (e.g. cvxopt, but there are also others).

Note that the runtime library versions of these packages (i.e. the non `-dev` versions, e.g. `libgmp10` and `libpng`) were removed in the switch from Alpine → Debian in ce07dadb5f6e2786d6ad9eabef00745b501ed425, which implies (b) and static linking.

One way to find out is to checkout 3917c1b and (after verifying you can build the image as-it-was), try building it without those deps and see where it fails.

IIRC, both Augur and Auspice had transitive dependencies on libpng at one point, though this may have since been loosened.

Generally speaking, GMP is often required for bignum support in numerical software, so it's not surprising to see. (I'm a little surprised we don't need it at runtime, but static linking could explain this. Slightly terrifyingly, I've run into libraries that make their use of a bignum library like GMP optional, falling back to a horrifyingly slow implementation when GMP isn't available. Hopefully that's not the case here!)

Dockerfile
```diff
@@ -136,6 +146,15 @@ RUN cd /nextstrain/auspice && npm update && npm install && npm run build && npm
 FROM python:3.7-slim-buster
 
 # Add system runtime deps
+# bzip2, gzip, xz-utils, zip, unzip, zstd: install compression tools
+# ca-certificates: ?
```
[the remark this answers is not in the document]
victorlin: Got it, noted in a9dd473.

Dockerfile
```diff
@@ -11,6 +11,16 @@ FROM python:3.7-slim-buster AS builder
 SHELL ["/bin/bash", "-c"]
 
 # Add system deps for building
+# autoconf, automake: for building VCFtools; may be used by package managers to build from source
+# build-essential: contains gcc, g++, make, etc. for building various tools; may be used by package managers to build from source
+# ca-certificates: ?
```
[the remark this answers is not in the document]
victorlin: Noted in a9dd473.

Dockerfile
victorlin: I've tested these on the current Dockerfile:

- `pkg-config` is needed by VCFtools. Noted in a9dd473.
- `libgmp-dev` is not needed anymore. Removed in 1bcd494.
- `libpng-dev` is not needed, but its dependency `zlib1g-dev` is needed by VCFtools. Updated in a47e358.

I checked out 3917c1b9aa0aa19e694f94f6bd3e56418f248269, but building from that commit no longer works due to changes in upstream URL availability. I did not dig deeper to find out exactly where `libgmp-dev` and `libpng-dev` were used.

victorlin: > I've run into libraries that make their use of a bignum library like GMP optional, falling back to a horrifyingly slow implementation when GMP isn't available. Hopefully that's not the case here!

That doesn't sound good! Maybe worth adding back in a separate PR and doing trial runs on the ncov workflow?

tsibley: Nice testing work.

I definitely think it's a good idea to confirm that GMP's presence/absence in the builder and/or final image doesn't affect execution times. ncov is probably a good test bench? or maybe monkeypox?

tsibley (approved):

tsibley:
Great to have this documentation!

PR #95, opened 2022-10-14
Update builder-scripts/ COPY command, consolidate other commands

victorlin:

See commit messages.

### Related issue(s)

_N/A_

### Testing

- [x] Checks pass


tsibley (approved):
Seems good to me. The `WORKDIR` directives will each add an additional layer and could be eliminated by using absolute paths (or an initial `cd`) in the `RUN` directives, but not sure if it matters much.

PR #96, opened 2022-10-17
Improve Node.js installation

victorlin:

According to the official ["Dockerfile best practices" guide](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/), it's a good idea to pair apt-get update with apt-get install in the same RUN statement.

This is mostly a future-proofing change, as the current script at https://deb.nodesource.com/setup_14.x already calls apt-get update. I think caching problems would only arise if (1) the Node.js version is updated, (2) the setup script no longer call apt-get update, and (3) the package install list is updated so that at least one package is missing from the APT cache pulled in the last RUN statement, which may be from an old Docker cache.

### Related issue(s)

_N/A_

### Testing

- [x] Checks pass


tsibley (approved):

jgadling:
NOTE -- this should fix the debian-bullseye build!

PR #66, opened 2022-07-05
chore: switch to Debian 11 (bullseye) from Debian 10 (buster)

corneliusroemer:
Debian 10 reaches EOL in August

This helps getting more up to date versions of things like `zstd` installed by `apt install`

Is there a reason we need to use Debian 10?

PR #100, opened 2022-10-18
Upgrade python & debian to the latest released version

jgadling:

Upgrade Python and Debian to the latest released version.

Python 3.7 is over four years old at this point. Our team has a variety of python scripts that we run alongside Nextstrain in Docker, and we'd like to be able to take advantage of the improved typing capabilities of Python 3.9+. In addition it upgrades to the latest debian LTS

### Related issue(s)

This supercedes #66 

### Testing

- [ ] Checks pass

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@
 # image with tools only needed during the image build.
 
 # First build the temporary image.
-FROM python:3.10.8-slim-buster AS builder
+FROM python:3.10-slim-buster AS builder
 
 # Execute subsequent RUN statements with bash for handy modern shell features.
 SHELL ["/bin/bash", "-c"]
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -5,7 +5,7 @@
 # image with tools only needed during the image build.
 
 # First build the temporary image.
-FROM python:3.7-slim-buster AS builder
+FROM python:3.10.8-slim-buster AS builder
```
tsibley, line 8: It'd be safe to make this just `3.10` since we didn't pin patch releases previously.

corneliusroemer:
Oh just saw this now and approved actions to run, thanks!

tsibley:
I'll look at this today, along with some of the other lingering PRs in this repo.

tsibley:
I addressed the Debian upgrade separately (#66) and reworked the changes here as #102. Will merge that when it passes CI.

PR #101, opened 2022-10-18
Split CI into separate jobs using artifact upload/download

### LINKED ISSUE #69: CI: Distinguish build/test/push phases

It would be good to clean up the increasingly complex `run` job in the CI workflow. Should also consider #68 here, since that would split the push phase to push `branch-…` before test phase, and push `latest` only after successful test.

victorlin:

Split CI to make the workflow file easier to follow.

### Related issue(s)

- Closes #69

### Testing

- [ ] Checks pass

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,7 @@
 name: CI
 on: [push, pull_request, workflow_dispatch]
 jobs:
-  run:
+  build:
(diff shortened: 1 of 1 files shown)


victorlin:
This should work, but it's slow - see https://github.com/nextstrain/docker-base/issues/69#issuecomment-1282965208. Closing because I don't think the improved readability is worth the extra run time at this point.

PR #102, opened 2022-10-19
Upgrade Python from 3.7 to 3.10

tsibley:
3.7 was chosen in early 2020, when 3.8 was only a few months old.  3.10 is now a year old, and we already use it in other Nextstrain contexts.

The motivation for this came from @jgadling who wanted to take advantage of newer Python typing features in scripts running alongside Nextstrain in this Docker image.

(Change made by @jgadling, commit written by @tsibley.)

(Rework of #100 to squash commits and write a commit message with more rationale.)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@
 # image with tools only needed during the image build.
 
 # First build the temporary image.
-FROM python:3.7-slim-bullseye AS builder
+FROM python:3.10-slim-bullseye AS builder
 
 # Execute subsequent RUN statements with bash for handy modern shell features.
 SHELL ["/bin/bash", "-c"]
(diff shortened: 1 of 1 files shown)


tsibley:
Verified both our ncov and monkeypox workflows (at least their slimmed down CI versions) work fine with this change (via `nextstrain/base:branch-upgrade-python`).

tsibley:
Our custom `nextstrain/ncov-ingest` image extending `nextstrain/base` will need some updating with this change, but I'll arrange that separately.

danrlu:
Thank you all!!

tsibley:
Noting since I just needed to know: this was first released as `nextstrain/base:build-20221019T225050Z`.

PR #103, opened 2022-10-21
Split CI into separate jobs using GitHub Container registry

### LINKED ISSUE #69: CI: Distinguish build/test/push phases

victorlin:

Split the single CI job to make the workflow file easier to follow. Note that this broadens the scope of the `devel/build` script to also push to a registry, but the overall local dev process is not hindered. See the README changes for more on that.

These changes were pulled out of #50 and repurposed. 6ac97e175dd6608a4987b63c13e5af4217934282 notes that the changes serve two purposes.

### Changes made outside of commits

1. Added @nextstrain-bot as an admin for the docker-base repository. Although the account is a member of @nextstrain/core which has write access to this repository, [write access does not include package deletion](https://docs.github.com/en/packages/learn-github-packages/configuring-a-packages-access-control-and-visibility#visibility-and-access-permissions-for-container-images). Only admin access allows package deletion.
2. Generated a personal access token for @nextstrain-bot named `org secret: GH_TOKEN_NEXTSTRAIN_BOT_MANAGE_PACKAGES` with `delete:packages`, `write:packages`, and other implicit scopes. Set to expire on 01/01/2023.
3. Created a new Actions secret in @nextstrain, [`GH_TOKEN_NEXTSTRAIN_BOT_MANAGE_PACKAGES`](https://github.com/organizations/nextstrain/settings/secrets/actions/GH_TOKEN_NEXTSTRAIN_BOT_MANAGE_PACKAGES), with the generated token as the value. Made it available to the docker-base repository.

### Related issue(s)

- Closes #69

### Testing

As of 02b53264a0b08fa1e673afae4eb7432b55e56284:

- [x] Checks pass
- [x] New concurrency limit + auto-cancel works ([new run](https://github.com/nextstrain/docker-base/actions/runs/3414216543), [canceled run](https://github.com/nextstrain/docker-base/actions/runs/3414243505))
- [x] https://github.com/nextstrain/docker-base/packages and https://github.com/orgs/nextstrain/packages get populated during CI runs, and are empty afterwards

### Post-merge

- [ ] `push-build` job runs as expected
- [ ] Verify that workflow concurrency is allowed on `master`

### Miscellaneous discussions

- [Consolidate `push-branch` and `push-build`](https://github.com/nextstrain/docker-base/pull/103#discussion_r1006172984)? Not possible.


devel/delete-from-ghcr
```diff
@@ -0,0 +1,68 @@
...
+    echo "Deleting the package version for $org/$container:$tag ..."
+    deleteVersionError=$(curl -s \
+        -X DELETE \
+        -H "Accept: application/vnd.github+json" \
+        -H "Authorization: Bearer $GITHUB_TOKEN" \
+        "https://api.github.com/orgs/$org/packages/container/$container/versions/$id" \
+    )
+    echo "$deleteVersionError"
+
+    if [[ "$deleteVersionError" =~ "You cannot delete the last tagged version of a package. You must delete the package instead." ]]; then
```
tsibley, line 47: Did you look into using `skopeo delete` instead of GitHub's REST API? I would think that might be less overhead and GitHub Package's container registry would handle this sort of nuance for us. (But maybe not!)

tsibley (commented):
I like this direction! I'm not sure about some of the details.

Are we going to have ephemeral `nextstrain/base` images showing up under https://github.com/orgs/nextstrain/packages (or https://github.com/nextstrain/docker-base/packages) when CI is running?

Is using ghcr.io for possibly-transient CI images going to make it awkward/difficult if we want to also publish the production images to ghcr.io in parallel to docker.io? ([as I alluded to elsewhere](https://github.com/nextstrain/docker-base/issues/69#issuecomment-1283139018))

devel/delete-from-ghcr
tsibley: If `skopeo delete` won't work, you might also consider https://github.com/actions/github-script to run a little JS program where Octokit is automatically available and setup.

devel/delete-from-ghcr
```diff
@@ -0,0 +1,68 @@
...
+        if [[ -n "$deletePackageError" ]]; then
+            echo "Could not delete $org/$container." >&2
+            exit 1
+        fi
+
+    elif [[ -n "$deleteVersionError" ]]; then
+        echo "Could not delete $org/$container:$tag." >&2
+        exit 1
+    fi
+done
```
victorlin, line 68: Re: [PR review comment](https://github.com/nextstrain/docker-base/pull/103#pullrequestreview-1157154143)

> Are we going to have ephemeral nextstrain/base images showing up under https://github.com/orgs/nextstrain/packages (or https://github.com/nextstrain/docker-base/packages) when CI is running?

Yes, but [the default visibility is private](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#pushing-container-images). Note that this visibility level is per-package, not package version (tag), so it's only useful if we want to keep all tags private.

> Is using ghcr.io for possibly-transient CI images going to make it awkward/difficult if we want to also publish the production images to ghcr.io in parallel to docker.io?

In this PR, I'm using the same image name/tag for both intermediate and destination registries. This won't work if the intermediate is the same as the destination, so either name or tag will have to change for the intermediate.

1. Changing the name: This would require some further parameterization of the `devel` scripts to specify image names, which are currently hardcoded as `nextstrain/base` and `nextstrain/base-builder`. A benefit would be that intermediate images will live in packages separate from final images, so intermediates can be kept private.
2. Changing the tag: This would require further expansion of the `copy-images` script to allow different src/dest tags, and intermediate images must be public.

I'm leaning towards (1), though I'm not worried about having public intermediate images if it comes to that.

.github/workflows/ci.yml
```diff
@@ -22,33 +20,99 @@ jobs:
...
-    - run: ./devel/build -r localhost:$REGISTRY_PORT -t $TAG
+    - run: ./devel/build -r ghcr.io -t $TAG
 
-    - if: github.event_name != 'pull_request'
-      uses: docker/login-action@v1
-      with:
-        username: ${{ secrets.DOCKER_LOGIN }}
-        password: ${{ secrets.DOCKER_PASSWORD }}
+    outputs:
+      tag: ${{ env.TAG }}
```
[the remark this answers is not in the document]
victorlin: Where did you see that? [I had to check](https://github.com/victorlin/github-test/actions/runs/3371745264/workflow), but yes you're right!

.github/workflows/ci.yml
```diff
@@ -22,33 +20,99 @@ jobs:
...
+    steps:
 
-    - if: github.event_name != 'pull_request' && startsWith(env.TAG, 'branch-')
-      name: Copy $TAG images to Docker Hub (non-default branch)
-      run: ./devel/copy-images -i localhost:$REGISTRY_PORT -o docker.io -t $TAG
+    - uses: docker/login-action@v1
+      with:
+        registry: ghcr.io
+        username: nextstrain-bot
+        password: ${{ secrets.GH_TOKEN_NEXTSTRAIN_BOT_MANAGE_PACKAGES }}
```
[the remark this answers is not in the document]
victorlin: Login is needed at least because the intermediate packages are currently private. I'm not sure how it would behave if it were public.

We shouldn't need `delete:packages` or `write:packages` for pulling, but I didn't want to create another token with only `read:packages`. That was mostly out of laziness, so if you think a separate token is better then I can do that.

.github/workflows/ci.yml
```diff
@@ -22,33 +20,99 @@ jobs:
...
         git clone https://github.com/nextstrain/zika-tutorial
         pip install nextstrain-cli
-        nextstrain build --image localhost:$REGISTRY_PORT/nextstrain/base:$TAG zika-tutorial -F
+        nextstrain build --image ghcr.io/nextstrain/base:${{ needs.build.outputs.tag }} zika-tutorial -F
 
-    - if: startsWith(env.TAG, 'build-')
-      name: Copy $TAG + latest images to Docker Hub (default branch)
+  push-branch:
+    if: startsWith(needs.build.outputs.tag, 'branch-') && github.event_name != 'pull_request'
+    needs: build
```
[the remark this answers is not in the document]
victorlin: I didn't think of ways, but thinking now and it seems difficult because only `push-branch` should run before `test`.

Having them separate provides a more intuitive UI page for the workflow run:

CI page

devel/delete-from-ghcr
victorlin: Thanks, I hadn't thought about about either of those! Will look into it.

.github/workflows/ci.yml
```diff
@@ -20,7 +22,9 @@ jobs:
       # and hyphens.
       run: echo "TAG=branch-${GITHUB_REF_NAME//[^A-Za-z0-9._-]/-}" | tee -a $GITHUB_ENV
 
-    - run: ./devel/build $TAG
+    - run: ./devel/start-localhost-registry $REGISTRY_PORT
```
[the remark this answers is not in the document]
victorlin: I did not, but yeah that looks to be per job, so wouldn't be useful here.

README.md
```diff
@@ -28,14 +28,24 @@ This is most helpful when you want the image to contain the latest version of a
...
+2. Build the image.
+
+    ```
+    ./devel/build
+    ```
+
+    By default, this tags the image with `latest` and pushes to `localhost:5000`. See instructions at the top of the script for additional options.
 
-By default, this tags the image with `latest`. If you want to use a custom tag
-name, you can specify it as the first argument to the script.
```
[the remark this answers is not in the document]
victorlin: Fixed in [force-push](https://github.com/nextstrain/docker-base/compare/662d7b2339acf2748df76bb6e47e3cb9f4338bf9..607d4d749a6c92070de4210bd9484a746584e967).

.github/workflows/ci.yml
```diff
@@ -22,33 +20,99 @@ jobs:
...
+
+    - run: mv ~/.docker/config.json ~/.docker/ghcr.io-config.json
+
+    - uses: docker/login-action@v1
+      with:
+        registry: docker.io
+        username: ${{ secrets.DOCKER_LOGIN }}
+        password: ${{ secrets.DOCKER_PASSWORD }}
+
+    - run: mv ~/.docker/config.json ~/.docker/docker.io-config.json
```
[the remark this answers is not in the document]
victorlin: Good thinking, it works! Done with [force-push](https://github.com/nextstrain/docker-base/compare/607d4d749a6c92070de4210bd9484a746584e967..82f7c379c1dd316d976eaa6745a09946b56d85d0).

devel/delete-from-ghcr
victorlin: `skopeo delete` [doesn't work on ghcr.io](https://github.com/nextstrain/docker-base/actions/runs/3373053143/jobs/5597277742). From [the docs](https://github.com/containers/skopeo/blob/62e698b5672236a3388afcf0a398fb0bb05f745b/docs/skopeo-delete.1.md):

> The effect of this is registry-specific; many registries don’t support this operation, or don’t allow it in some circumstances / configurations.

victorlin: `actions/github-script` is neat! Switched over to it in [force-push](https://github.com/nextstrain/docker-base/compare/b0e7395451b3cf316fec99a028e5b267a462a76e..2cb9afeafb9049319977d8793f955dfea8340ff5).

devel/delete-from-ghcr
```diff
@@ -0,0 +1,68 @@
...
+
+    # Each tag should only correspond to one package version.
+    # Pushing an existing tag will untag the existing version and add the tag
+    # to the newly pushed version.
+    id=$(curl -s \
+          -H "Accept: application/vnd.github+json" \
+          -H "Authorization: Bearer $GITHUB_TOKEN" \
+          "https://api.github.com/orgs/$org/packages/container/$container/versions" \
+        | jq --arg TAG $tag '[ .[] | select(.metadata.container.tags[] == $TAG ) ] | .[0].id' \
+    )
```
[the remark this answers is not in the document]
victorlin: This isn't necessary after switching to `actions/github-script`.

.github/workflows/ci.yml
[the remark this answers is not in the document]
tsibley: I didn't see it documented, but the fact that this works (at least, I think it must or this workflow wouldn't work in your testing, right?) means it does.

FWIW, I think your test workflow is testing something different, as it's using shell interpolation with an environment variable, not workflow interpolation (`${{ }}`) with the `env` context.

.github/workflows/ci.yml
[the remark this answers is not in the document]
tsibley: Nope, I think that's fine.

.github/workflows/ci.yml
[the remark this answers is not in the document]
tsibley: Nod. I _think_ we could make a single `push` job with a dynamically-set `needs:` and thus avoid the "one of the push-* jobs never runs" behaviour, but it's just fine as-is with two jobs too.

.github/workflows/ci.yml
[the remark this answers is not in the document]
tsibley: :tada:

devel/delete-from-ghcr
tsibley: :tada:

devel/delete-from-ghcr
tsibley: Nod. Ok. Let's cross that bridge when we get to it then. Thanks for the answers!

devel/delete-from-ghcr.js
```diff
@@ -0,0 +1,57 @@
+// Delete Nextstrain images with a specific tag from the GitHub Container registry.
+//
+// This file is intended for use by the GitHub action "actions/github-script".
```
tsibley, line 3: I guess using `actions/github-script` makes it harder to test/perform locally. Ah well.

devel/delete-from-ghcr
tsibley: I was kinda hoping Octokit had more sugar over the package tag to version id lookup and deletion of the last version stuff, but [alas it doesn't](https://octokit.github.io/rest.js/v19#packages-delete-package-version-for-org).

.github/workflows/ci.yml
```diff
@@ -1,5 +1,11 @@
 name: CI
 on: [push, pull_request, workflow_dispatch]
+
+# Prevent intermediate images from being used by another run.
+concurrency:
+  group: ${{ github.ref }}
+  cancel-in-progress: true
```
tsibley, line 7: Hmm, I've tinkered with `concurrency` before, and two things stand out here:

1. I'm not sure we want to apply `cancel-in-progress` to runs on `master` (e.g. releases). Doing so will mean that two independent pushes to master (potentially separated by some time if workflow run times are long) which should ideally produce two different `build-*` tags will instead produce just the latest one. When "push to master" means "make a release", I think the relationship should stay 1:1. I think fine to cancel in progress jobs on branches and PRs though, since these are meant for testing/dev only.
2. Concurrency groups apply repo-wide, i.e. across workflows. It's a good idea to explicitly scope the group key to this workflow (e.g. `${{ github.workflow }}@${{ github.ref }}` or similar), since that's an assumed limitation here.

victorlin: I've never used this feature before, so thanks for those notes!

It should be fine to disable concurrency for just the default branch given independent pushes should produce unique `build-*` tags (near-zero chance of collision since the value is time-based). I'll look into that.

.github/workflows/ci.yml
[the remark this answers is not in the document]
victorlin: I played with this a bit and don't think it's possible to dynamically set `needs`: https://github.com/community/community/discussions/38492

.github/workflows/ci.yml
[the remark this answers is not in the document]
victorlin: I think I'm misunderstanding. Whether the evaluation of `${{ env.TAG }}` comes during or after the job step should be irrelevant for `outputs` which comes at the end, right?

But yeah, it works so this doesn't matter much!

devel/delete-from-ghcr.js
victorlin: [This force-push](https://github.com/nextstrain/docker-base/compare/c2d069f17aaa0a830bb02ea93fb80cfdf1d72970..68d46254213d48c8d93b8b2032527729e682ea0a) adds a note on how to use it locally if you wish. Note that this whole pushing intermediates to ghcr.io thing is CI-specific and not used during local dev – that uses a local registry which can be cleaned up using `devel/stop-localhost-registry`.

.github/workflows/ci.yml
victorlin: [This force-push](https://github.com/nextstrain/docker-base/compare/68d46254213d48c8d93b8b2032527729e682ea0a..02b53264a0b08fa1e673afae4eb7432b55e56284) should address the above points. I did not test it, but added an entry in the PR description's post-merge tasks.

tsibley (approved):

PR #104, opened 2022-10-26
Add `evofr` for forecasts

joverlee521:
Creating PR to start some general discussions:

1. Should `evofr` be added to docker-base or should we create a separate forecasts image? 
I've added it here to easily test with the `nextstrain/base:branch-evofr` image, but I'm not sure if we should actually add it to docker-base since it's such a niche tool. We are currently only using `evofr` for [forecasts-ncov](https://github.com/nextstrain/forecasts-ncov), but we will probably expand to other pathogens. It's unclear to me how much it will get used by outside Nextstrain users. 

1. How often should the image get updated? 
Since `evofr` is still in development, would it make sense to add a GH Action to the `evofr` repo to build the image on each push to the main branch?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -192,6 +192,9 @@ WORKDIR /nextstrain/auspice
 RUN /builder-scripts/download-repo https://github.com/nextstrain/auspice release . \
  && npm update && npm install && npm run build && npm link
 
+# Add evofr for forecasting
+RUN pip3 install evofr
+
 # ———————————————————————————————————————————————————————————————————— #
 
 # Now build the final image.


trvrb:
1. I'd personally benefit from this as a single image. I'd view `evofr` as another component of the Nextstrain toolkit, just like `augur` and `nextclade`. If `forecasts-ncov` is a baseline Nextstrain "pathogen" repo (which I think it is), then I think it makes sense to have core components of this as part of base image. Going forward, we're going to have "forecasting" workflows (primarily using `evofr`) and "phylodynamics" workflows (primarily using `augur`).

2. I'd treat rebuilding the same way we do with `fauna` and just pull `main` (https://github.com/nextstrain/docker-base/blob/master/Dockerfile#L173). I wouldn't add automated image rebuilds for each push to `main` and instead update add hock (but assume that updates for augur, auspice, etc... will keep things pretty current). `evofr` may move to a release model as well and we could change then.

Dockerfile
```diff
@@ -145,6 +145,10 @@ RUN pip3 install git+https://github.com/cov-lineages/constellations.git@v0.1.1
 RUN pip3 install git+https://github.com/cov-lineages/pango-designation.git@19d9a537b9
 RUN pip3 install pysam==0.19.1
 
+# Add evofr for forecasting
+# Not available on PyPI, so install via git URLs
+# Currently not pinning to any commit yet since it's still in active development
+RUN pip3 install git+https://github.com/blab/evofr.git
```
victorlin, line 151: This should come after `ARG CACHE_DATE` otherwise an older version will stay in the cache indefinitely.

Currently, only tools under the "Nextstrain components" section come after `CACHE_DATE`, since we want to have the latest version of all Nextstrain components (either latest release or latest on default branch) without the hassle of updating pins. I'm not familiar with `evofr` but if it's part of the Nextstrain toolkit as @trvrb describes, then it would make sense to put it in that section anyways. Otherwise the section could be renamed to "Nextstrain components and unpinned programs".

victorlin (changes requested):

victorlin:
1. Adding to `nextstrain/base` seems much easier in the short-term, and we can always remove it later in favor of alternatives if desired.
2. I think this is mostly up to whoever plans to use `evofr`, so I don't have a strong opinion here. If "in development" means possible broken state on the main branch, then no. Otherwise, automating makes sense since all the other Nextstrain components here are unpinned and use either the latest release or latest from default branch if not versioned.

Dockerfile
joverlee521: Ah, thank you! I had no idea how Docker build caches work.

tsibley:
I'm in complete concurrence with @victorlin here.

tsibley:
> `evofr` may move to a release model as well and we could change then.

Is this something we could just do ~now and then trigger and install from released versions here? Actual PyPI releases will also be important for producing equivalent Conda packages.

evofr is already using Poetry and successfully produces a source dist and wheel with `poetry build`. The version has seen several bumps over time. It seems the only step missing is running `poetry publish` when appropriate. That could easily be manual for now, with automation coming later.

This is mostly a question for @marlinfiggins, I suppose.

Dockerfile
victorlin: After a good month of working on this repo, I understand it more now but still not 100%!

joverlee521:
Marking this as ready for review since the general consensus is to just add `evofr` to docker-base. 
I'm fine with manually running the GH Action for pushing updates to Docker if there are any upstream `evofr` changes that we want to pull in.

victorlin (approved):

tsibley (approved):
Ok with me to merge, though I think there are still two questions worth answering:

- [x] [Is the `main` branch of evofr considered always stable](https://github.com/nextstrain/docker-base/pull/104#issuecomment-1292626023)? Or does it ever see commits which render it broken/impaired?
- [ ] [Would Marlin (or someone else) be up for making releases to PyPI](https://github.com/nextstrain/docker-base/pull/104#issuecomment-1293883708), e.g. when the version is bumped?

I've [pinged](https://bedfordlab.slack.com/archives/C7SDVPBLZ/p1667538723038609) @marlinfiggins in Slack.

- [ ] We also need to account for evofr in https://github.com/nextstrain/conda-base. This would be best done by re-packaging it for Conda, likely as part of Bioconda.

tsibley:
[Per in Slack](https://bedfordlab.slack.com/archives/C7SDVPBLZ/p1667538999005969?thread_ts=1667538723.038609&cid=C7SDVPBLZ):

1. `main` should always be stable.
2. @marlinfiggins is happy to make releases! (but would love some help too)

tsibley:
I've pushed tags for all versions in git and published the latest version (0.1.12) to PyPI: https://pypi.org/project/evofr/

PR #105, opened 2022-11-08
Uncache Nextstrain CLI

joverlee521:
Move the Nextstrain CLI installation to below the CACHE_DATE so that it does not stay in the cache indefinitely. I believe this was the intention when we stopped pinning the Nextstrain CLI to a specific version.

### Testing

- [x] Checks pass


victorlin (approved):

tsibley:
Thanks for catching this, @joverlee521!

PR #110, opened 2022-12-03
Add NCBI Datasets CLI

joverlee521:

The NCBI Datasets command line tools will be used to fetch virus packages¹ for various pathogen ingest pipelines.

The `datasets` command fetches the virus packages and the `dataformat` command transforms the nested JSON data structure to flat TSV metadata files.

¹ https://www.ncbi.nlm.nih.gov/datasets/docs/v2/reference-docs/data-packages/virus-genome/

### Testing

- [x] Checks pass

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -197,6 +197,10 @@ RUN /builder-scripts/download-repo https://github.com/nextstrain/auspice release
 # Add evofr for forecasting
 RUN pip3 install evofr
 
+# Add NCBI Datasets command line tools for access to NCBI Datsets Virus Data Packages
+RUN curl -fsSL -o /final/bin/datasets https://ftp.ncbi.nlm.nih.gov/pub/datasets/command-line/v2/linux-amd64/datasets
+RUN curl -fsSL -o /final/bin/dataformat https://ftp.ncbi.nlm.nih.gov/pub/datasets/command-line/v2/linux-amd64/dataformat
+
 # ———————————————————————————————————————————————————————————————————— #
 
 # Now build the final image.


Dockerfile
```diff
@@ -197,6 +197,10 @@ RUN /builder-scripts/download-repo https://github.com/nextstrain/auspice release
 # Add evofr for forecasting
 RUN pip3 install evofr
 
+# Add NCBI Datasets command line tools for access to NCBI Datsets Virus Data Packages
+RUN curl -fsSL -o /final/bin/datasets https://ftp.ncbi.nlm.nih.gov/pub/datasets/command-line/v2/linux-amd64/datasets
+RUN curl -fsSL -o /final/bin/dataformat https://ftp.ncbi.nlm.nih.gov/pub/datasets/command-line/v2/linux-amd64/dataformat
```
tsibley, line 202: Let's prefix both of these with `ncbi-`, e.g. to get `ncbi-datasets`. An unscoped `datasets` command seems bound to be confusing.

Dockerfile
```diff
@@ -145,7 +145,7 @@ RUN pip3 install git+https://github.com/cov-lineages/pango-designation.git@19d9a
 RUN pip3 install pysam==0.19.1
 
 
-# 4. Add Nextstrain components
+# 4. Add Nextstrain components and unpinned programs
```
tsibley, line 148: This will need a rebase onto current `master` to avoid a conflicting edit.

tsibley (changes requested):

tsibley:
We also need to arrange for these to be added to our other runtimes:
- Conda via https://github.com/nextstrain/conda-base
- Ambient install instructions?

Dockerfile
joverlee521: Ah yup, I can see how an unscoped `datasets` command can be confusing within the Nextstrain context. 
Is it also possible to also prefix with `ncbi-` in conda-base?

joverlee521: Or do you mean to do something similar to Nextclade 2 like so: 

```Suggestion
# Add NCBI Datasets command line tools for access to NCBI Datsets Virus Data Packages
RUN curl -fsSL -o /final/bin/ncbi-datasets https://ftp.ncbi.nlm.nih.gov/pub/datasets/command-line/v2/linux-amd64/datasets \
 && ln -sv datasets /final/bin/ncbi-datasets
RUN curl -fsSL -o /final/bin/ncbi-dataformat https://ftp.ncbi.nlm.nih.gov/pub/datasets/command-line/v2/linux-amd64/dataformat \
 && ln -sv dataformat /final/bin/ncbi-dataformat
```

tsibley: Ah, I meant just supporting e.g. `ncbi-datasets`, not `datasets` too.

But the question about Conda is a good one. Hmm. I assumed we were going to have to package them ourselves and thus could make that choice, but I see now that there is an `ncbi-datasets-cli` package. Despite the package name being qualified, it installs the unqualified `datasets` and `dataformat` commands. (I find it ridiculous that NCBI thinks it's ok to use such generic names!)

I think we either:

- Accept `datasets` and `dataformat` for all runtimes
- Repackage NCBI Datasets' CLI ourselves for Conda with the qualified binary names

joverlee521: I lean towards keeping `datasets` and `dataformat` as is so that our uses of them follow the NCBI docs and also less for us to maintain 🤷

tsibley: Yeah, agreed, though I'll cringe when we run into our first issue with the generic names… :see_no_evil:

tsibley:
> We also need to arrange for these to be added to our other runtimes:

Oh jeez, hadn't seen https://github.com/nextstrain/conda-base/pull/5 yet.

Dockerfile
joverlee521: I guess we'll see 🤞

joverlee521:
Manually triggered the [ncov-ingest update-image workflow](https://github.com/nextstrain/ncov-ingest/actions/runs/3651952181) to pull in the latest nextstrain/base.

PR #118, opened 2022-12-12
Upgrade pangolin to v4

jgadling:

Upgrade pangolin (with UShER support) to version 4.1.3

I've done this by creating another intermediate-build image for UShER. The UShER team *is* publishing an image to dockerhub [here](https://registry.hub.docker.com/r/pathogengenomics/usher), but it's based on an OS with different shared-library versions than the one we're using, so it's difficult/impractical to just use the built binaries from the upstream image.

I can easily refactor this to skip adding a third upstream image repository if that's a problem.

### Testing

I spot-checked that pangolin works in UShER and scorpio mode in the final image.

#### UShER mode:
```
nextstrain:/tmp $ pangolin test_genome.fasta
****
Pangolin running in usher mode.
****
<content omitted>

Using UShER as inference engine.
****
Output file written to: /tmp/lineage_report.csv
```
#### PangoLEARN mode
```
nextstrain:/tmp $ pangolin --analysis-mode fast test_genome.fasta
****
Pangolin running in pangolearn mode.
****
Warning: pangoLEARN mode may use a significant amount of RAM, be aware that it will not suit every system.
Maximum ambiguity allowed is 0.3.
****
Query file:     /tmp/test_genome.fasta
****
<content omitted>
Processing block of 3 sequences 12/12/2022, 18:03:03
Complete 12/12/2022, 18:03:03
****
Output file written to: /tmp/lineage_report.csv
```

- [ ] Checks pass

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,6 +31,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         jq \
         libsqlite3-dev \
         pkg-config \
+        rsync \
         zlib1g-dev
 
 # Install a specific Node.js version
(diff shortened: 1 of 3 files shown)


tsibley:
@jgadling Thanks for this proposal. I've held off on reviewing/testing this because I think we'd rather drop Pangolin/PangoLEARN all together if we can—rather than upgrade it—per discussion started in https://github.com/nextstrain/conda-base/issues/7. We haven't resolved that discussion yet, but we should try to do so one way or another soon.

victorlin:
Closing this since pangolin is no longer in the Docker image as of #237.

PR #120, opened 2022-12-16
Remove redundant Auspice build

victorlin:

Auspice has a `prepare` script that runs the `build` script automatically after `npm install`. This means the explicit call to run the `build` script is redundant and can be removed to improve build times.

### Testing

- [x] Checks pass
- [ ] `nextstrain view --docker --image nextstrain/base:branch-victorlin-remove-redundant-auspice-build` works

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -250,7 +250,7 @@ RUN /builder-scripts/download-repo https://github.com/nextstrain/augur "$(/build
 # used for the same reasons described above.
 WORKDIR /nextstrain/auspice
 RUN /builder-scripts/download-repo https://github.com/nextstrain/auspice release . \
- && npm update && npm install && npm run build && npm link
+ && npm update && npm install && npm link
 
 # Add evofr for forecasting
 RUN pip3 install evofr


victorlin:
CI passes, but `nextstrain view` is unable to find `'/nextstrain/auspice/dist'` and fails.

I think this has to do with the upgrade to Node.js 16. Compare 1afd07dae89be39b145ab465d42f5fd8b12cedd8 (which introduces the Node.js 16 upgrade) and the previous commit f8ea4dabd93dcefd19f3888bb229dfd71b73b650. Node.js 14 logs only show one occurrence of `node auspice.js build --verbose` while Node.js 16 logs show two (one being redundant).

Both [NPM v7](https://docs.npmjs.com/cli/v7/using-npm/scripts) (packaged with Node.js 14) and [NPM v8](https://docs.npmjs.com/cli/v8/using-npm/scripts) (packaged with Node.js 16) docs say that the `prepare` script should run on local `npm install` (and this is what I've observed on my local machine), so I'm not sure why there is a difference in behavior here. But given what's observed, this change should be bundled with the upgrade to Node.js 16.

PR #121, opened 2022-12-20
Update Docker Hub login

victorlin:

Use new token and remove unnecessary encryption of username.

### Related issue(s)

- https://github.com/nextstrain/private/issues/73

### Testing

- [x] Checks pass


victorlin:
Merging without review since this is a trivial change and checks pass, which conforms that the new token works.

PR #122, opened 2022-12-20
Upgrade to Node.js 16

victorlin:

Upgrade to Node.js 16 and remove a redundant Auspice build that [comes with](https://github.com/nextstrain/docker-base/pull/120#issuecomment-1360218169) the upgrade.

### Pre-merge tasks

- Merge [#98](https://github.com/nextstrain/docker-base/pull/98) which is the target branch of this PR. Otherwise without cross-compilation, Auspice installs/builds are [much slower](https://github.com/nextstrain/docker-base/pull/50#issuecomment-1354247750).

### Testing

- [x] Checks pass

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -214,7 +214,7 @@ RUN curl -fsSL -o /final/bin/nextclade2 https://github.com/nextstrain/nextclade/
 # ⁴ https://github.com/mapbox/node-pre-gyp/blob/v1.0.10/lib/node-pre-gyp.js#L186
 WORKDIR /nextstrain/auspice
 RUN /builder-scripts/download-repo https://github.com/nextstrain/auspice release . \
- && npm update && npm install && npm run build && npm link
+ && npm update && npm install && npm link
 
 # Add NCBI Datasets command line tools for access to NCBI Datsets Virus Data Packages
 RUN curl -fsSL -o /final/bin/datasets https://ftp.ncbi.nlm.nih.gov/pub/datasets/command-line/v2/linux-${TARGETARCH}/datasets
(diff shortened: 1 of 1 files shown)


jameshadfield:
Any reason not to jump straight to node v18 (which is also LTS)?

With regards to possibly removing the redundant auspice build step (#98, which indicates it may be rolled into this PR) I would favour the more explicit approach of  `npm install --ignore-scripts ...` and a later `npm run build`, especially if the latter will be followed by linking.

victorlin:
@jameshadfield

> Any reason not to jump straight to node v18 (which is also LTS)?

Auspice doesn't officially support v18 (but will soon). Also, v16 is what's currently used on [nextstrain.org](https://github.com/nextstrain/nextstrain.org/blob/2685e6b363cf70808fa11ee38ac520dfb3a9cdc4/package.json#L11) and [auspice.us](https://github.com/nextstrain/auspice.us/blob/0708c367333b1bf9d1b3616ca144b376a9139025/package.json#L7).

After https://github.com/nextstrain/auspice/pull/1560 is released, we can coordinate an upgrade of v16 → v18 for this Docker image, the two deployed apps, and individual dev environments.

jameshadfield:
P.S. Node 16 reaches end-of-life [in September this year](https://github.com/nodejs/release#release-schedule) so I'd suggest going straight to node 18 in this PR, as [auspice supports it](https://github.com/nextstrain/auspice/pull/1560) even though we still use v16 on nextstrain.org & auspice.us.

victorlin:
I'm going to leave this PR as 14→16 and make another PR for 16→18. Just in case there are issues with 14→18, it'll be easier to figure out where the issue comes from.

I plan to merge this once CI passes and I verify ability to run `auspice view` on the test image.

PR #124, opened 2022-12-22
Update shell usage

victorlin:

See commit messages and updates to README.

### Related issue(s)

N/A

### Testing

- [x] Checks pass


tsibley (approved):
LGTM. Even though CI passed, I'd hazard we might still see unanticipated failures due to the added strictures of `-euo pipefail`, but we can fix those if they arise.

PR #125, opened 2022-12-22
Add new directory for scripts used in final stage

victorlin:

This de-clutters the top-level directory and makes it more navigable, as all scripts can be categorized by usage:

1. Used in the builder stage (builder-scripts/)
2. Used in the final stage (final-scripts/)
3. Used in the development process (devel/)

### Related issue(s)

N/A

### Testing

- [x] Checks pass


tsibley (commented):
Frankly, I'm pretty meh on this and don't really understand the motivation. I don't think it's no good _per se_, but I don't see the point at all.

> This de-clutters the top-level directory and makes it more navigable

Is the top-level really that cluttered with 7 `ls`-visible items? Does reducing that to 5 items really make it more navigable? I don't think that increased nesting always equates to less clutter/greater navigability and generally try to use minimum nesting.

These items, `bashrc`, `entrypoint`, and `entrypoint-aws-batch`, are _essential parts_ of the image itself, not build-time helpers like `builder-scripts/` or `devel/` so I don't see it paralleling those. I don't see them as the same things as the other "scripts". The essential parts of a repo like this seem like they should be at the top-level to me.

IIRC, `builder-scripts/` were also initially co-located in `devel/` before being split out somewhat recently, and I think I was kinda meh about that too. So I don't see "it parallels builder-scripts/" as a strong rationale.

If the top-level really feels too cluttered to you, I would be less meh about a proposal to move these source image contents into a directory structure that mirrors their destination, e.g. something like:

```
bashrc → src/etc/bash.bashrc
entrypoint → src/sbin/entrypoint
entrypoint-aws-batch → src/sbin/entrypoint-aws-batch
```

…but I'd still think that nesting is not necessary.

victorlin:
@tsibley Clutter is not my main concern, it's more about having a self-documenting repo structure.

When I first started working on this repo, it wasn't clear to me that there were two types of scripts in `devel/`: (1) used by builder and (2) used by CI. Only when modifying the files and seeing [usage](https://github.com/nextstrain/docker-base/blame/0aaaaa7b7c63ad013e09e3e7e1f563d5fee87051/Dockerfile#L83-L84) did I understand the separation.

Up until recently, I glanced over the 3 top-level files that are moved in this PR since I haven't had to work with them directly. If they were in a directory named `final-scripts/`, I would at least have understood that they are used in the final image.

I can understand if all Dockerfile-related files are in the top-level directory (i.e. `mv builder-scripts/* .`), but right now the repo structure feels split between that and this PR.

tsibley:
**The top line is ok to merge this.** That's fine even though there's disagreement. It's minor!

The rest of this is only FYI explaining my position.

> I can understand if all Dockerfile-related files are in the top-level directory (i.e. `mv builder-scripts/* .`)…

Mm, my point wasn't to put all Dockerfile-related things in the top-level. It was that things in the repo should be separated by "stuff that makes it into the images users run" (e.g. the contents of the final image) and "stuff that doesn't" (e.g. stuff used only in the process of producing the image). `builder-scripts/`, for example, is stuff that doesn't.

My perspective here is that the multiple build stages and the build process and what not are all _development_ concerns. These things exist solely because of how we're accomplishing the image builds at a given point in time and could plausibly be replaced with an entirely different approach to produce the ~same image (e.g. like the immense effort you've just done to support arm64!).

So to me, the organizational split that's nice to have for someone browsing the repo is "this is development stuff you don't have to care about unless you're developing" and "this is stuff that's in the image you use, which you probably care about more".

`final-scripts/` doesn't read to me as "these things are essential parts of the image". And esp. if I know that `builder-scripts/` are only build-time helpers (for the builder image), I'm inclined to think `final-scripts/` are also only build-time helpers (for the final image). But they're not.

> Clutter is not my main concern, it's more about having a self-documenting repo structure.

I'd encourage you to include main concerns like "a self-documenting repo" in the commit rationale instead of only including "clutter". :upside_down_face:

victorlin:
> So to me, the organizational split that's nice to have for someone browsing the repo is "this is development stuff you don't have to care about unless you're developing" and "this is stuff that's in the image you use, which you probably care about more".

Thanks for adding this perspective. When I created this PR, I was deep in the developer mindset without recognizing the benefits of the current placement for non-developers. Looking back now, I feel more meh on this and will close it since there are no strong votes for it.

> I'd encourage you to include main concerns like "a self-documenting repo" in the commit rationale instead of only including "clutter". 🙃

Of course! I hadn't been able to find the right words at the time of writing but your rebuttal helped me articulate better. I'd update the commit if it were to be merged.

tsibley:
> Of course! I hadn't been able to find the right words at the time of writing but your rebuttal helped me articulate better. I'd update the commit if it were to be merged.

:raised_hands: :heart:

PR #126, opened 2022-12-23
Add libgomp1 for running FastTree

### LINKED ISSUE #123: FastTreeDblMP doesn't work in docker
> I tried running `augur tree --alignment whatever.fasta --method fasttree` and got the following error:

```text
FastTreeDblMP: error while loading shared libraries: libgomp.so.1: cannot open shared object file: No such file or directory
```

_Originally posted by @fanninpm in https://github.com/nextstrain/docker-base/issues/97#issuecomment-1363004029_

____

```text
$ docker image ls --digests nextstrain/base
REPOSITORY          TAG                      DIGEST                                                                    IMAGE ID            CREATED             SIZE
nextstrain/base     build-20221214T204556Z   sha256:e341aadab5681050ddc751218b254b1172276d87516466103db84399fa808d54   84e33405645c        7 days ago          2.41GB
```

victorlin:

As far as I can tell from the official software page¹, no dependencies are documented.

However, there are (unofficial?) Linux packages²³ that list libc6 and libgomp1 as dependencies. Adding the latter for now since `augur tree --method fasttree` fails with the error:

    FastTreeDblMP: error while loading shared libraries: libgomp.so.1: cannot open shared object file: No such file or directory

¹ http://www.microbesonline.org/fasttree/
² https://packages.ubuntu.com/focal/fasttree
³ https://packages.debian.org/sid/science/fasttree

### Related issue(s)

- Fixes #123

### Testing

- [x] Checks pass
- [x] `docker run nextstrain/base:branch-victorlin-fix-fasttree FastTreeDblMP -h` shows the help output instead of the error in the original issue.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -271,6 +271,7 @@ FROM python:3.10-slim-bullseye AS final
 # dos2unix: tsv-utils needs unix line endings
 # jq: may be used by workflows
 # less: for usability in an interactive prompt
+# libgomp1: for running FastTree
 # libsqlite3: for pyfastx (for Augur)
 # perl: for running VCFtools
 # ruby: may be used by workflows
@@ -285,6 +286,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         gzip \
         jq \
         less \
+        libgomp1 \
         libsqlite3-0 \
         perl \
         ruby \


victorlin:
Merging without review since this is a small change and testing was successful.

Issue #128, opened 2023-02-21
Different Augur versions between amd64 and arm64 images


victorlin:
### Current Behavior

It is possible that images of the same tag for different architectures have different Augur versions.

### Expected behavior

Images of the same tag for different architectures should be guaranteed to have the same Augur version.

### How to reproduce

Not sure that this is easily reproducible, but it was brought up by @corneliusroemer [in Slack](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1676906669414509) for the Augur 21.0.0 → 21.0.1 update.

In the [triggered image rebuild](https://github.com/nextstrain/docker-base/actions/runs/4206926139/jobs/7301049323), the two images pulled different versions of Augur:

```
2023-02-17T19:29:48.6526581Z #90 [linux/amd64 builder 42/47] RUN /builder-scripts/download-repo https://github.com/nextstrain/augur "$(/builder-scripts/latest-augur-release-tag)" .  && pip3 install --editable .
… nextstrain-augur==21.0.0

2023-02-17T19:30:52.5529403Z #99 [linux/arm64 builder 42/47] RUN /builder-scripts/download-repo https://github.com/nextstrain/augur "$(/builder-scripts/latest-augur-release-tag)" .  && pip3 install --editable .
… nextstrain-augur==21.0.1
```

A [relevant line](https://github.com/nextstrain/augur/actions/runs/4206823143/jobs/7300812344#step:10:54) from the Augur 21.0.1 release job logs shows when the new version was published to PyPI:

```
2023-02-17T19:28:11.7789440Z https://pypi.org/project/nextstrain-augur/21.0.1/
```

My speculation is that [the PyPI endpoint used to retrieve the version](https://github.com/nextstrain/docker-base/blob/d59452c4800c09e7f418c80d70379523a7cf84ef/builder-scripts/latest-augur-release-tag#LL11C12-L11C55) was only updated sometime within the 1 minute difference between the two invocations of `latest-augur-release-tag`.

### Possible solution

1. Determine the latest Augur version using the latest GitHub Release or by parsing `__version__.py`, instead of a PyPI endpoint that has some lag time.
2. Only trigger the Docker image rebuild after the new Augur version is known to be available at the PyPI endpoint used to retrieve the version.
3. Add tests to verify that Augur versions (and other software versions) are the same before publishing the public final images.
[...]

PR #130, opened 2023-02-28
CI: Validate images across platforms

victorlin:

Start off by checking software in which versions are known to be time-varying.

### Related issue(s)

- Addresses the main point in #128.

### Testing

- [x] Checks pass

Tested locally:

1. With a tag known to have differing Augur versions.

    <details>

    <summary>
    Commands and output
    </summary>

    ```
    % ./devel/validate-platforms -r docker.io -t build-20230217T192833Z
    [linux/amd64] Pulling image...
    docker.io/nextstrain/base:build-20230217T192833Z
    [linux/amd64] Checking that the platform is expected...
    [linux/amd64] Generating report file: /var/folders/rk/s2p6nv1d13lcf0dynjkvxyl80000gp/T/validate-platforms-XXXXXX.WEPtnBnz/linux/amd64
    [linux/amd64] Determining software versions...
    [linux/arm64] Pulling image...
    docker.io/nextstrain/base:build-20230217T192833Z
    [linux/arm64] Checking that the platform is expected...
    [linux/arm64] Generating report file: /var/folders/rk/s2p6nv1d13lcf0dynjkvxyl80000gp/T/validate-platforms-XXXXXX.WEPtnBnz/linux/arm64
    [linux/arm64] Determining software versions...
    The report for linux/amd64 has the following contents:
    $ nextstrain --version
    nextstrain.cli 6.1.0.post1
    $ nextalign --version
    nextalign 2.11.0
    $ nextclade --version
    nextclade 2.11.0
    $ augur --version
    augur 21.0.0
    $ auspice --version
    2.43.0
    $ python3 -c "from importlib.metadata import version; print(version(\"evofr\"))"
    0.1.16
    $ datasets --version
    datasets version: 14.13.0
    $ dataformat version
    14.13.0
    $ python3 -c "from importlib.metadata import version; print(version(\"phylo-treetime\"))"
    0.9.5
    Comparing against other platforms...
    --- linux/amd64	2023-03-10 12:13:04.000000000 -0800
    +++ linux/arm64	2023-03-10 12:13:11.000000000 -0800
    @@ -7,3 +7,3 @@
     $ augur --version
    -augur 21.0.0
    +augur 21.0.1
     $ auspice --version
    Failure!
    % echo $?
    1
    ```

    </details>

2. With a tag known to have the same Augur version.

    <details>

    <summary>
    Commands and output
    </summary>

    ```
    % ./devel/validate-platforms -r docker.io -t build-20230220T161818Z
    [linux/amd64] Pulling image...
    docker.io/nextstrain/base:build-20230220T161818Z
    [linux/amd64] Checking that the platform is expected...
    [linux/amd64] Generating report file: /var/folders/rk/s2p6nv1d13lcf0dynjkvxyl80000gp/T/validate-platforms-XXXXXX.OHXS3kXH/linux/amd64
    [linux/amd64] Determining software versions...
    [linux/arm64] Pulling image...
    docker.io/nextstrain/base:build-20230220T161818Z
    [linux/arm64] Checking that the platform is expected...
    [linux/arm64] Generating report file: /var/folders/rk/s2p6nv1d13lcf0dynjkvxyl80000gp/T/validate-platforms-XXXXXX.OHXS3kXH/linux/arm64
    [linux/arm64] Determining software versions...
    The report for linux/amd64 has the following contents:
    $ nextstrain --version
    nextstrain.cli 6.1.0.post1
    $ nextalign --version
    nextalign 2.11.0
    $ nextclade --version
    nextclade 2.11.0
    $ augur --version
    augur 21.0.1
    $ auspice --version
    2.43.0
    $ python3 -c "from importlib.metadata import version; print(version(\"evofr\"))"
    0.1.16
    $ datasets --version
    datasets version: 14.13.0
    $ dataformat version
    14.13.0
    $ python3 -c "from importlib.metadata import version; print(version(\"phylo-treetime\"))"
    0.9.5
    Comparing against other platforms...
    Success!  All versions the same.
    % echo $?
    0
    ```

    </details>

### Post-merge

- [ ] Create issue for checking diff of `pip list` without error-ing

### DIFF
--- a/devel/validate-platforms
+++ b/devel/validate-platforms
@@ -1,0 +1,110 @@
+#!/bin/bash
+#
+# Validate different platform builds of the final Nextstrain image.
+#
+set -euo pipefail
+
+# Set default values.
+registry=localhost:5000
+tag=latest
+
+# Read command-line arguments.
+while getopts "r:t:" opt; do
+    case "$opt" in
+        r) registry="$OPTARG";;
+        t) tag="$OPTARG";;
+        *) echo "Usage: $0 [-r <registry>] [-t <tag>]" 1>&2; exit 1;;
+    esac
+done
+
+IMAGE="$registry/nextstrain/base:$tag"
+PLATFORMS=(linux/amd64 linux/arm64)
+
+main() {
+    # Check that every platform image got the same versions of important (e.g.
+    # first-party) software for which we don't pin a specific version (e.g. we
+    # install whatever the latest version is at build time).
+
+    local report_dir
+    report_dir="$(mktemp -dt "$(basename "$0")"-XXXXXX)"
+
+    for platform in "${PLATFORMS[@]}"; do
+        echo "[$platform] Pulling image..."
+        docker pull -q --platform "$platform" "$IMAGE"
+
+        echo "[$platform] Checking that the platform is expected..."
+        check-platform "$platform"
+
+        # Initialize a directory for the report file, ensuring slashes in the
+        # platform name are subdirs.
+        report="$report_dir/$platform"
+        echo "[$platform] Generating report file: $report"
+        mkdir -p "$(dirname "$report")"
+
+        # Create a report file for the platform.
+        # This should include all software below ARG CACHE_DATE in the Dockerfile
+        # in addition to other important software.
+        echo "[$platform] Determining software versions..."
+        docker-run "$platform" bash -c '
+            PS4="\$ "
+            set -x
+
+            nextstrain --version
+            nextalign --version
+            nextclade --version
+            augur --version
+            auspice --version
+            python3 -c "from importlib.metadata import version; print(version(\"evofr\"))"
+            datasets --version
+            dataformat version
+
+            python3 -c "from importlib.metadata import version; print(version(\"phylo-treetime\"))"
+        ' >"$report" 2>&1
+    done
+
+    # Compare contents of the first platform's report file against others.
+    first_report="$report_dir/${PLATFORMS[0]}"
+    echo "The report for ${PLATFORMS[0]} has the following contents:"
+    cat "$first_report"
+
+    echo "Comparing against other platforms..."
+    # NOTE: if running on macOS ≥13, you may need to install GNU diff for the
+    # --from-file option.
+    if cd "$report_dir" && diff --unified=1 --from-file="${PLATFORMS[0]}" "${PLATFORMS[@]:1}"; then
+        echo "Success!  All versions the same." >&2
+    else
+        echo "Failure!" >&2
+        exit 1
+     fi
+}
+
+check-platform() {
+    # Check that the platform is actually what we expect it to be.
+    local platform="$1"
+
+    python_platform_string="$(docker-run "$platform" python -c "import platform; print(platform.platform())")"
+
+    case "$platform" in
+        linux/amd64)
+            if [[ "$python_platform_string" != *"x86_64"* ]]; then
+                echo "Platform $platform not detected." 1>&2; exit 1
+            fi;;
+        linux/arm64)
+            if [[ "$python_platform_string" != *"aarch64"* ]]; then
+                echo "Platform $platform not detected." 1>&2; exit 1
+            fi;;
+        *)
+            echo "Platform $platform not supported." 1>&2; exit 1;;
+    esac
+}
+
+docker-run() {
+    # Run a command under the final Nextstrain image built for a specific
+    # platform.
+    local platform="$1"
+    local command=("${@:2}")
+
+    docker run --rm --platform "$platform" "$IMAGE" "${command[@]}"
+}
+
+main
--- a/devel/validate-platforms
+++ b/devel/validate-platforms
@@ -46,8 +46,10 @@ main() {
         # in addition to other important software.
         echo "[$platform] Determining software versions..."
         docker-run "$platform" bash -c '
-            PS4="\$ "
-            set -x
+            function echo-command {
+                echo "$ $BASH_COMMAND"
+            }
+            trap echo-command DEBUG
 
             nextstrain --version
             nextalign --version
@@ -59,7 +61,7 @@ main() {
             dataformat version
 
             python3 -c "from importlib.metadata import version; print(version(\"phylo-treetime\"))"
-        ' >"$report" 2>&1
+        ' >"$report"
     done
 
     # Compare contents of the first platform's report file against others.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,6 +82,24 @@ jobs:
       env:
         DOCKER_DEFAULT_PLATFORM: ${{ matrix.platform }}
 
+  validate-platforms:
+    name: Validate platforms
+    needs: build
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+
+    - uses: docker/setup-qemu-action@v2
+
+    - uses: docker/login-action@v2
+      with:
+        registry: ghcr.io
+        username: nextstrain-bot
+        password: ${{ secrets.GH_TOKEN_NEXTSTRAIN_BOT_MANAGE_PACKAGES }}
+
+    - name: Validate final images
+      run: ./devel/validate-platforms -r ghcr.io -t ${{ needs.build.outputs.tag }}
+
   # "Push" (copy) the builder and final images from GitHub Container Registry to
   # Docker Hub, where they will persist. Do this regardless of test results.
   push-branch:
@@ -110,7 +128,7 @@ jobs:
   # Docker Hub, where they will persist. Only do this if tests pass.
   push-build:
     if: startsWith(needs.build.outputs.tag, 'build-')
-    needs: [build, test]
+    needs: [build, test, validate-platforms]
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
@@ -134,7 +152,7 @@ jobs:
   # Delete the builder and final images from GitHub Container Registry.
   cleanup-registry:
     if: always()
-    needs: [build, test, push-branch, push-build]
+    needs: [build, test, validate-platforms, push-branch, push-build]
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
--- a/Dockerfile
+++ b/Dockerfile
@@ -206,6 +206,8 @@ RUN pip3 install pysam==0.19.1
 
 # Allow caching to be avoided from here on out by calling
 # docker build --build-arg CACHE_DATE="$(date)"
+# NOTE: All versioned software added below should be checked in
+# devel/validate-platforms.
 ARG CACHE_DATE
 
 # Install our own CLI so builds can do things like `nextstrain deploy`
--- a/README.md
+++ b/README.md
@@ -68,6 +68,14 @@ variable to a new timestamp first:
 Otherwise, letting the build process use the cached layers will save you time
 during development iterations.
 
+### Validate the images
+
+Before using the images, they should be checked for any inconsistencies.
+
+    ./devel/validate-platforms
+
+The output and exit code will tell you whether validation is successful.
+
 ### Using the images locally
 
 Since the images are pushed directly to the local registry, they are not


tsibley (commented):
I think it'd be more useful to take a different approach when comparing the versions. Instead of constructing an array for the versions of each component and comparing them individually, how about constructing a single text file report for each platform and then `diff`-ing them? Combined with meaningful report file names, this would provide more readily interpretable output and also provide all version differences at once instead of stopping at the first mismatch. The code would also be quite a bit reduced and simpler.

Thoughts?

victorlin:
@tsibley nice, that sounds like a better approach! I'll update this PR with that.

tsibley:
@victorlin Sounds good. I tweaked what was there earlier this morning into this while tinkering with it:

```diff
diff --git a/devel/validate-platforms b/devel/validate-platforms
index 804434c..6fefc85 100755
--- a/devel/validate-platforms
+++ b/devel/validate-platforms
@@ -18,23 +18,18 @@ while getopts "r:t:" opt; do
 done
 
 IMAGE="$registry/nextstrain/base:$tag"
+PLATFORMS=(linux/amd64 linux/arm64)
 
 main() {
-    compare-versions
-}
-
-compare-versions() {
     # Check software where downloaded versions are not deterministic at
     # different points in time (e.g. if "latest" is downloaded).
+    local output
+    output="$(mktemp -dt "$(basename "$0")"-XXXXXX)"
 
-    nextclade_versions=()
-    augur_versions=()
-    auspice_versions=()
-    evofr_versions=()
-    datasets_versions=()
-    dataformat_versions=()
+    for platform in "${PLATFORMS[@]}"; do
+        # Ensure slashes in platforms are subdirs
+        mkdir -p "$(dirname "$output/$platform")"
 
-    for platform in linux/amd64 linux/arm64; do
         echo "[$platform] Pulling image..."
         docker pull -q --platform "$platform" "$IMAGE"
 
@@ -42,59 +37,27 @@ compare-versions() {
         check-platform "$platform"
 
         echo "[$platform] Determining software versions..."
-        nextclade_versions+=("$(run-command "$platform" nextclade --version)")
-        augur_versions+=("$(run-command "$platform" augur --version)")
-        auspice_versions+=("$(run-command "$platform" auspice --version)")
-        evofr_versions+=("$(run-command "$platform" pip show evofr)")
-        datasets_versions+=("$(run-command "$platform" datasets --version)")
-        dataformat_versions+=("$(run-command "$platform" dataformat version)")
+        run-command "$platform" bash -c '
+            PS4="\$ "
+            set -x
+
+            nextclade --version
+            augur --version
+            auspice --version
+            pip show evofr
+            datasets --version
+            dataformat version
+        ' >"$output/$platform" 2>&1
+        cat "$output/$platform"
     done
 
-    echo "Comparing Nextclade versions..."
-    if different-values "${nextclade_versions[@]}"; then
-        echo "Nextclade versions differ!" 1>&2; exit 1
+    echo "Diffing versions..."
+    if cd "$output" && diff --unified --from-file="${PLATFORMS[0]}" "${PLATFORMS[@]:1}"; then
+        echo "Success!  All versions the same." >&2
+    else
+        echo "Failure!" >&2
+        exit 1
     fi
-
-    echo "Comparing Augur versions..."
-    if different-values "${augur_versions[@]}"; then
-        echo "Augur versions differ!" 1>&2; exit 1
-    fi
-
-    echo "Comparing Auspice versions..."
-    if different-values "${auspice_versions[@]}"; then
-        echo "Auspice versions differ!" 1>&2; exit 1
-    fi
-
-    echo "Comparing evofr versions..."
-    if different-values "${evofr_versions[@]}"; then
-        echo "evofr versions differ!" 1>&2; exit 1
-    fi
-
-    echo "Comparing NCBI Datasets (datasets) versions..."
-    if different-values "${datasets_versions[@]}"; then
-        echo "NCBI Datasets (datasets) versions differ!" 1>&2; exit 1
-    fi
-
-    echo "Comparing NCBI Datasets (dataformat) versions..."
-    if different-values "${dataformat_versions[@]}"; then
-        echo "NCBI Datasets (dataformat) versions differ!" 1>&2; exit 1
-    fi
-
-    echo "Success!"
-}
-
-different-values() {
-    local values=("$@")
-
-    first_value="${values[0]}"
-    for value in "${values[@]}"; do
-        if [[ "$first_value" != "$value" ]]; then
-            echo "Values \"$first_value\" and \"$value\" are different." 1>&2
-            return 0 # true
-        fi
-    done
-    echo "Values are all \"$first_value\"."
-    return 1 # false
 }
 
 check-platform() {
```

victorlin:
@tsibley great, I'm going to use some of that.

Also, I just noticed that the OS I'm using (macOS Ventura) comes with a version of `diff` that doesn't support `--from-file`:

```
% diff --version
Apple diff (based on FreeBSD diff)
% diff --from-file
diff: unrecognized option `--from-file'
```

Apparently this is [new as of macOS Ventura](https://github.com/NixOS/nix/issues/7286#issuecomment-1311223003). Bummer.

corneliusroemer:
> EDIT: I installed GNU diff via Homebrew's [diffutils](https://formulae.brew.sh/formula/diffutils), much better.
Ha, I must have already installed gnu diff a while ago since I never got that error. The way to go would be to run the script in a docker container 🙃 Though I'm not sure it's possible to run docker in docker.

victorlin:
@corneliusroemer I imagine this will mostly be run on the GitHub Actions runner, which should continue to have GNU diff. However, it would be good to consider local dev on common OS's (latest macOS included), so for now I added a note (rather than automatically supporting 2 versions of `diff`):

https://github.com/nextstrain/docker-base/blob/468cb918971f161fce0f9df54c34770614a2f9ee/devel/validate-platforms#L62-L63

corneliusroemer:
> I imagine this will mostly be run on the GitHub Actions runner, which should continue to have GNU diff. However, it would be good to consider local dev on common OS's (latest macOS included), so for now I added a note (rather than automatically supporting 2 versions of diff):

Great! Re local dev experience, that's why I'd suggest changing default to point at public docker registry. It should work out of the box. Could be useful for debugging etc. Do you have a docker registry set up locally or why use localhost:5000 as default for the registry?

devel/validate-platforms
```diff
@@ -0,0 +1,101 @@
...
+
+    # Compare contents of the first platform's report file against others.
+    first_report="$report_dir/${PLATFORMS[0]}"
+    echo "The report for ${PLATFORMS[0]} has the following contents:"
+    cat "$first_report"
+
+    echo "Comparing against other platforms..."
+    # NOTE: if running on macOS ≥13, you may need to install GNU diff for the
+    # --from-file option.
+    if cd "$report_dir" && diff --unified=1 --from-file="${PLATFORMS[0]}" "${PLATFORMS[@]:1}"; then
```
tsibley, line 64: :+1: This output is nicer than what I had. :-)

devel/validate-platforms
```diff
@@ -0,0 +1,101 @@
...
+        docker-run "$platform" bash -c '
+            PS4="\$ "
+            set -x
+
+            nextclade --version
+            augur --version
+            auspice --version
+            pip show evofr | grep Version
+            datasets --version
+            dataformat version
```
tsibley, line 52: _non-blocking_

Why report `datasets` and `dataformat`? I'd think we'd mostly (maybe even only) want to ensure that our first-party software is identical.

_[a couple minutes later]_

Ah, if the reason is because they're unpinned in our Dockerfile, then perhaps include a comment about this rationale here and there so we know a) why and b) to keep it synced?

devel/validate-platforms
```diff
@@ -0,0 +1,101 @@
...
+        # Create a report file for the platform.
+        echo "[$platform] Determining software versions..."
+        docker-run "$platform" bash -c '
+            PS4="\$ "
+            set -x
+
+            nextclade --version
+            augur --version
+            auspice --version
+            pip show evofr | grep Version
```
tsibley, line 50: Include `nextalign` and `nextstrain` too? They're also unpinned, first-party tools.

tsibley, line 50: _non-blocking_

As an alternative to `pip show | grep`, we could more directly inspect the version:

```
python3 -c 'from importlib.metadata import version; print(version("evofr"))'
```

Quicker too.

devel/validate-platforms
```diff
@@ -0,0 +1,101 @@
...
+
+IMAGE="$registry/nextstrain/base:$tag"
+PLATFORMS=(linux/amd64 linux/arm64)
+
+main() {
+    # Check software where downloaded versions are not deterministic at
+    # different points in time (e.g. if "latest" is downloaded).
+
+    local report_dir
+    report_dir="$(mktemp -dt "$(basename "$0")")"
```
tsibley, line 28: _blocking_

```suggestion
    report_dir="$(mktemp -dt "$(basename "$0")"-XXXXXX)"
```

The X's in my patch were intentional; they're a required part of the template for GNU coreutils' mktemp:

```
$ ./devel/validate-platforms -r docker.io
mktemp: too few X's in template ‘validate-platforms’
```

from mktemp(1):

> TEMPLATE must contain at least 3 consecutive 'X's in last component.

tsibley (changes requested):

devel/validate-platforms
```diff
@@ -0,0 +1,118 @@
...
+main() {
+    report_names=()
+
+    version_commands=(
+        "nextclade --version"
+        "augur --version"
+        "auspice --version"
+        "pip show evofr | grep Version"
+        "datasets --version"
+        "dataformat version"
```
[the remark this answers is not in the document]
tsibley: Ah, I don't think we want to ensure exact identity of third-party software versions between our platform images. This may not even be possible due to dependency resolution solutions varying between platforms. Even if it is, it'll be a continual headache, and not really the point of this image.

tsibley:
> Apparently this is [new as of macOS Ventura](https://github.com/NixOS/nix/issues/7286#issuecomment-1311223003). Bummer.

Ha. Burned by OS versions, because I'd actually checked that `diff` on `blab-mac` had `--from-file` and was surprised to find it was GNU (but from 2002, presumably due to licensing issues like bash).

devel/validate-platforms
victorlin: I noticed when CI failed... looks like another [difference between macOS and other OS's](https://unix.stackexchange.com/q/30091/181911) 😕

On a GitHub Codespace instance:

```
$ uname -a
Linux codespaces-cf4b1b 5.4.0-1103-azure #109~18.04.1-Ubuntu SMP Wed Jan 25 20:53:00 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

$ mktemp --version
mktemp (GNU coreutils) 8.30
Copyright (C) 2018 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <https://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Jim Meyering and Eric Blake.

$ man mktemp | tail -n 1
GNU coreutils 8.30                   September 2019                            MKTEMP(1)

$ mktemp -dt "validate-platforms-XXX"
/tmp/validate-platforms-Ege
```

On macOS Ventura:

```
% mktemp --version
mktemp: illegal option -- -
usage: mktemp [-d] [-q] [-t prefix] [-u] template ...
       mktemp [-d] [-q] [-u] -t prefix 

% man mktemp | tail -n 1
macOS 13.1                      August 11, 2014                     macOS 13.1

% mktemp -dt "validate-platforms-XXX"
/var/folders/m5/30jq1dcn439_lzq16twm_jcc0000gn/T/validate-platforms-XXX.bdPORZe5
```

The macOS version doesn't replace the X's but it still creates a valid temp directory, so I'll add them back in.

Curious, any reason you chose 6 X's over the minimum 3?

tsibley: > The macOS version doesn't replace the X's but it still creates a valid temp directory, so I'll add them back in.

Yep. It doesn't replace the Xs in `-t prefix` because it treats that as a "prefix" (to which it prepends the temp dir and appends the random chars) as opposed to a "template", e.g. the difference between these two usages:

```
usage: mktemp [-d] [-q] [-t prefix] [-u] template ...
       mktemp [-d] [-q] [-u] -t prefix 
```

> Curious, any reason you chose 6 X's over the minimum 3?

Habit because it's more secure, though that's hardly a concern in this usage.

devel/validate-platforms
victorlin: I like this! It also avoids the weirdly formatted output due to piping, which I didn't bother fixing to make it look right:

```
$ pip show evofr
$ grep Version
Version: 0.1.16
```

devel/validate-platforms
victorlin: The comment here already existed, but maybe it could be worded more clearly?

https://github.com/nextstrain/docker-base/blob/468cb918971f161fce0f9df54c34770614a2f9ee/devel/validate-platforms#L24-L25

The newest force-push includes a comment in the Dockerfile:

https://github.com/nextstrain/docker-base/blob/c70aa6c2d37aed5572ae20f1fdfbd0c287559bb7/Dockerfile#L207-L211

devel/validate-platforms
victorlin: Updated in [force-push](https://github.com/nextstrain/docker-base/compare/468cb918971f161fce0f9df54c34770614a2f9ee..c70aa6c2d37aed5572ae20f1fdfbd0c287559bb7).

victorlin: Good catch. Added in [force-push](https://github.com/nextstrain/docker-base/compare/468cb918971f161fce0f9df54c34770614a2f9ee..c70aa6c2d37aed5572ae20f1fdfbd0c287559bb7).

devel/validate-platforms
victorlin: Took the best of both sides!

devel/validate-platforms
```diff
@@ -0,0 +1,118 @@
+#!/bin/bash
+#
+# Validate different platform builds of the final Nextstrain image.
+#
+set -euo pipefail
+
+# Set default values.
+registry=localhost:5000
```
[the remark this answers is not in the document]
victorlin: The [force-push](https://github.com/nextstrain/docker-base/compare/468cb918971f161fce0f9df54c34770614a2f9ee..c70aa6c2d37aed5572ae20f1fdfbd0c287559bb7) adds a snippet in the README to make this more apparent.

I manually tested with `-r docker.io` since that's the easiest way to showcase a reproducible success and failure. Going forwards, I'd expect usage to be primarily CI (where `-r ghcr.io` is used for many other scripts), and the occasional local development (where the localhost default would be consistently convenient).

devel/validate-platforms
victorlin: Good catch. Added in [force-push](https://github.com/nextstrain/docker-base/compare/468cb918971f161fce0f9df54c34770614a2f9ee..c70aa6c2d37aed5572ae20f1fdfbd0c287559bb7).

devel/validate-platforms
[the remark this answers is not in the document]
victorlin: Good to note the inconsistency here, but I also agree that it isn't necessary to validate against. Maybe it would be useful to print out the diff of `pip list` in case we want an easy way to look back at it on a particular tag. Is that what you were thinking @corneliusroemer?

devel/validate-platforms
victorlin: X's added in [force-push](https://github.com/nextstrain/docker-base/compare/468cb918971f161fce0f9df54c34770614a2f9ee..c70aa6c2d37aed5572ae20f1fdfbd0c287559bb7).

devel/validate-platforms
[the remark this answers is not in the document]
corneliusroemer: We definitely want to know about differences - there are not going to be many and they are very easy to check.

I agree we don't want to fail on these by default, but they're very useful to have available at a glance. This makes it _much_ easier to check whether there are differences and also just generally see which versions we have installed.

This would have helped debugging the bug that started all this _a lot_ (remember the biopython -> bcbiogff -> augur shebang).

While we may not want to fail for the entire dependency chain, I think it would be useful to ensure consistency of an important subset.

Wrote up a suggestion here: https://github.com/nextstrain/docker-base/pull/132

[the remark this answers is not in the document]
victorlin: Added small comments to #132. I was thinking something different but can be added independently of your proposal. Instead of failing on the diff of `pip list`, we can print it out separately to easily look back on the subtle differences such as `cryptography` package version differences between the two platform builds.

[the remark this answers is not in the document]
corneliusroemer: Sure, either works. If we want to fail on few, we can just check in one style for simplicity. And make the full pip list pure info.

devel/validate-platforms
```diff
@@ -0,0 +1,105 @@
...
+        echo "[$platform] Determining software versions..."
+        docker-run "$platform" bash -c '
+            PS4="\$ "
+            set -x
+
+            nextstrain --version
+            nextalign --version
+            nextclade --version
+            augur --version
+            auspice --version
```
corneliusroemer, line 53:
```suggestion
            auspice --version
            treetime --version
```

We definitely need to add `treetime` here - it's at the centre of much of `augur`.

The reason it may not be in this list is because it was created based on the way we install tools (e.g. binary vs pip or pypi) but that's not how it should be. But at least from my perspective what matters is what's in there, not how it ended up there.

victorlin: The current list is 1:1 with the tools under the Dockerfile section "Add unpinned programs and Nextstrain components" (see https://github.com/nextstrain/docker-base/pull/130#discussion_r1127163040).

I don't disagree with adding `treetime`, but want to be clear on moving away from the current criteria here and defining a new one.

Maybe the criteria can be updated to something like: include all tools under that section, plus any others that are worth tracking (slightly subjective, but better to err on the side of caution)? The "others" would include `treetime` and the extras you've proposed in #132.

corneliusroemer:
Let's merge this sooner than later? I don't see it doing any harm merged as is. If it fails, we fix it.

devel/validate-platforms
victorlin: The current section title is misleading – I'm updating it in 6da76965daa9c74e7b0defed0fca52140e2d6c13. Doesn't matter if it's "Nextstrain" or not, the original criterion here is to keep 1:1 with unpinned tools.

[Like you said](https://github.com/nextstrain/docker-base/pull/130#issuecomment-1462167413), I think this can be merged as-is (pending proper review). More checks can be added later.

corneliusroemer: Since there's no harm in adding a pinned tool let's add tree time as we may well unpin.

And is tree time really pinned? Or does it have minimum version.

Tests pass, so for me review passes, it can always be refactored later, this is tooling and it's not a huge issue if it churns a bit.

devel/validate-platforms
tsibley: _non-blocking_

Yeah, this comment wasn't clear to me:

https://github.com/nextstrain/docker-base/blob/468cb918971f161fce0f9df54c34770614a2f9ee/devel/validate-platforms#L24-L25

I think I was thrown off by "latest", which in this context I was thinking about as an image tag. "Not deterministic at different points in time" is not super clear to me. (It's deterministic at different points in time, but time-varying. No?)

Maybe a reword to something like?

> Check that every platform image got the same versions of important (e.g. first-party) software where we don't pin a specific version (e.g. we install whatever the latest version is at build time).

devel/validate-platforms
tsibley: I have no issue adding `treetime --version` to the list (with a comment).

devel/validate-platforms
victorlin: Oh yes, "time-varying" is what I meant. Reworded in [force-push](https://github.com/nextstrain/docker-base/compare/c70aa6c2d37aed5572ae20f1fdfbd0c287559bb7..7d8e9480ba91af55f0a225718b405f63570a465e).

devel/validate-platforms
victorlin: Added treetime in [force-push](https://github.com/nextstrain/docker-base/compare/c70aa6c2d37aed5572ae20f1fdfbd0c287559bb7..7d8e9480ba91af55f0a225718b405f63570a465e).

Note that `treetime` is not installed as a command (maybe that should change), so it is checked by:

```sh
python3 -c "from importlib.metadata import version; print(version(\"phylo-treetime\"))"
```

devel/validate-platforms
```diff
@@ -46,8 +46,10 @@ main() {
         # in addition to other important software.
         echo "[$platform] Determining software versions..."
         docker-run "$platform" bash -c '
-            PS4="\$ "
-            set -x
+            function echo-command {
+                echo "$ $BASH_COMMAND"
+            }
+            trap echo-command DEBUG
```
tsibley: _non-blocking_

I don't understand the reason for this change, even after reading the good commit message:

>  Use trap instead of set -x
>
> This allows errors such as "command not found" to terminate the script.
>
>  I considered BASH_XTRACEFD but that isn't available on bash <4.0, which
> is where the default bash version of macOS is tied to indefinitely.

Why does `set -x` not allow errors such as "command not found" to terminate the script?

(I was also thinking, "Do we _want_ those errors to terminate it? or should we let them pass and get picked up in the diff later?", but realized the edge case there is some command being missing from all platform images and thus producing no diff.)

tsibley: Also, FWIW, a more minimal but equivalent trap would be:

```bash
trap 'echo "$ $BASH_COMMAND"' DEBUG
```

victorlin: Oh, maybe the commit message could be better!

> Why does `set -x` not allow errors such as "command not found" to terminate the script?

Because `set -x` outputs to `stderr` by default, so there's no way to differentiate the outputted commands from actual errors when redirecting `stderr` to the report. This can be changed by setting `BASH_XTRACEFD` but I went against that for reasons in the commit message.

victorlin: > ```shell
> trap 'echo "$ $BASH_COMMAND"' DEBUG
> ```

I was trying to use something like this but couldn't figure out how to get it to work properly while inside the `bash -c '…'` literal string. I just found a workaround though:

```
bash -c $'
    trap \'echo "$ $BASH_COMMAND"\' DEBUG
    …
' >"$report"
```

devel/validate-platforms
corneliusroemer: Thanks! Do I understand correctly that it's not available as command because it's not installed explicitly, only as dependency through augur?
Don't see why we shouldn't install treetime explicitly - it's very useful to have.

victorlin: > Do I understand correctly that it's not available as command because it's not installed explicitly, only as dependency through augur?

Yes, see #139

devel/validate-platforms
victorlin: > realized the edge case there is some command being missing from all platform images and thus producing no diff.

Yes, I realized when I tried `treetime --version` and noticed the unusual behavior.

tsibley: _continues to be non-blocking_

> Because `set -x` outputs to `stderr` by default, so there's no way to differentiate the outputted commands from actual errors when redirecting `stderr` to the report. This can be changed by setting `BASH_XTRACEFD` but I went against that for reasons in the commit message.

Oh. Hmm. Another way without using `BASH_XTRACEFD` is to

    set -x
    exec 2>&1

if you want to merge stdout and stderr.

But I think I'm still missing something, because it went from using `set -x` to stderr with:

    docker run … >report 2>&1

to `trap` + `echo` to stdout with:

    docker run … >report

Where those producing different interleavings of lines in `report`? If so, that's probably a Docker issue around how it handles stdio ("unusually", is the answer), or potentially with the `docker run` invocation in this PR.

OH! Nevermind, I finally get that you _want_ to separate stdout and stderr, but want the command tracing to go to stdout. Got it. Sorry for being slow on the uptake here.

But speaking of the `docker run` invocation, I just noticed something there… will comment separately.

> but couldn't figure out how to get it to work properly while inside the `bash -c '…'` literal string. I just found a workaround though:

You probably already are, but if not, beware that `$'…'` strings have other semantic changes as well, such as replacing C-style escapes like `\t` with a literal tab.

The more typical way to do it without `$'…'` is to use different kinds of quoted strings adjacent to each other (as adjacent quoted strings are treated as one; Python does something similar):

```bash
bash -c '
    trap '"'"'echo "$ $BASH_COMMAND"'"'"' DEBUG
    …
' >"$report"
```

This is admittedly very ugly and you have to carefully count/match quotes when reading/writing, but it's idiomatic and readily identified on sight as an area of "here be dragons".

devel/validate-platforms
```diff
@@ -0,0 +1,112 @@
...
+    esac
+}
+
+docker-run() {
+    # Run a command under the final Nextstrain image built for a specific
+    # platform.
+    local platform="$1"
+    local command=("${@:2}")
+
+    docker run --platform "$platform" "$IMAGE" "${command[@]}"
```
tsibley, line 109: This `docker run` invocation wants a `--rm` flag so it doesn't leave a bajillion stopped containers around when run locally. (In CI I guess it doesn't matter.)

tsibley (changes requested):

devel/validate-platforms
tsibley: (Thanks for being patient with me!)

tsibley: (The `function echo-command` approach also remains just fine. No need for a change here, hence _non-blocking_, I just wanted to understand.)

victorlin: Ok, I'm glad we are on the same page now! Thanks for taking the time to understand. I'll update the description of 7d8e9480ba91af55f0a225718b405f63570a465e with the other alternatives, but keep the actual changes as-is.

devel/validate-platforms
victorlin: Fixed in force-push.

devel/validate-platforms
victorlin: Reworded in b8135dbd719112f513a5a5747737efd55a7315e0.

victorlin:
Noting here that the script did it's job in [this CI run](https://github.com/nextstrain/docker-base/actions/runs/5272308356/jobs/9535666699) where NCBI datasets [15.6.0 was released](https://github.com/ncbi/datasets/commit/065c62144fec17c547fbdb67e445b50337bafcb3) between the time it was fetched in the `linux/amd64` build and the time it was fetched in the `linux/arm64` build.

PR #132, opened 2023-03-07
compromise proposal: use some of pip list

corneliusroemer:
PR to suggest how to use subset of `pip list`

Not a real PR, but easiest way to propose difference, I think.

The pip list output is nice, better than the "command newline version_number" pattern

```
[linux/amd64] Pulling image...
registry.hub.docker.com/nextstrain/base:latest
[linux/amd64] Checking that the platform is expected...
[linux/amd64] Generating report file: /var/folders/qf/4kkcfypx0gbfb0t9336522_r0000gn/T/validate-platforms-7FIoow/linux/amd64
[linux/amd64] Determining software versions...
[linux/arm64] Pulling image...
registry.hub.docker.com/nextstrain/base:latest
[linux/arm64] Checking that the platform is expected...
[linux/arm64] Generating report file: /var/folders/qf/4kkcfypx0gbfb0t9336522_r0000gn/T/validate-platforms-7FIoow/linux/arm64
[linux/arm64] Determining software versions...
The report for linux/amd64 has the following contents:
$ nextalign --version
nextalign 2.12.0
$ nextclade --version
nextclade 2.12.0
$ auspice --version
2.44.0
$ pip list
$ grep -E 'bcbio-gff|biopython|cycler|evofr|nextstrain-augur|nextstrain-cli|numpy|numpyro|pandas|phylo-treetime|pyfastx|pysam|rethinkdb|scikit-learn|scipy|snakemake|xopen'
bcbio-gff                0.6.9
biopython                1.80
cycler                   0.11.0
evofr                    0.1.16
nextstrain-augur         21.0.1      /nextstrain/augur
nextstrain-cli           6.2.0
numpy                    1.24.2
numpyro                  0.9.2
pandas                   1.5.3
phylo-treetime           0.9.5
pyfastx                  0.8.4
pysam                    0.19.1
rethinkdb                2.4.9
scikit-learn             1.2.1
scipy                    1.10.1
snakemake                5.10.0
xopen                    1.7.0
Comparing against other platforms...
Success!  All versions the same.
```

### DIFF
--- a/devel/validate-platforms
+++ b/devel/validate-platforms
@@ -40,20 +40,41 @@ main() {
         echo "[$platform] Generating report file: $report"
         mkdir -p "$(dirname "$report")"
 
+        # Versions to track from `pip list`
+        tracked_packages=(bcbio-gff
+            biopython
+            cycler
+            evofr
+            nextstrain-augur
+            nextstrain-cli
+            numpy
+            numpyro
+            pandas
+            phylo-treetime
+            pyfastx
+            pysam
+            rethinkdb
+            scikit-learn
+            scipy
+            snakemake
+            xopen
+        )
+        tracked_packages_string=${tracked_packages[*]}
+        tracked_packages_delim=${tracked_packages_string// /|}
+
         # Create a report file for the platform.
         echo "[$platform] Determining software versions..."
         docker-run "$platform" bash -c '
             PS4="\$ "
             set -x
 
-            nextstrain --version
             nextalign --version
             nextclade --version
-            augur --version
             auspice --version
-            python3 -c "from importlib.metadata import version; print(version(\"evofr\"))"
-            datasets --version
-            dataformat version
+            # datasets --version
+            # dataformat version
+
+            pip list 2>&1 | grep -E "'"$tracked_packages_delim"'"
         ' >"$report" 2>&1
     done
 
@@ -70,7 +91,7 @@ main() {
     else
         echo "Failure!" >&2
         exit 1
-     fi
+    fi
 }
 
 check-platform() {
--- a/devel/validate-platforms
+++ b/devel/validate-platforms
@@ -1,0 +1,105 @@
+#!/bin/bash
+#
+# Validate different platform builds of the final Nextstrain image.
+#
+set -euo pipefail
+
+# Set default values.
+registry=localhost:5000
+tag=latest
+
+# Read command-line arguments.
+while getopts "r:t:" opt; do
+    case "$opt" in
+        r) registry="$OPTARG";;
+        t) tag="$OPTARG";;
+        *) echo "Usage: $0 [-r <registry>] [-t <tag>]" 1>&2; exit 1;;
+    esac
+done
+
+IMAGE="$registry/nextstrain/base:$tag"
+PLATFORMS=(linux/amd64 linux/arm64)
+
+main() {
+    # Check software where downloaded versions are not deterministic at
+    # different points in time (e.g. if "latest" is downloaded).
+
+    local report_dir
+    report_dir="$(mktemp -dt "$(basename "$0")"-XXXXXX)"
+
+    for platform in "${PLATFORMS[@]}"; do
+        echo "[$platform] Pulling image..."
+        docker pull -q --platform "$platform" "$IMAGE"
+
+        echo "[$platform] Checking that the platform is expected..."
+        check-platform "$platform"
+
+        # Initialize a directory for the report file, ensuring slashes in the
+        # platform name are subdirs.
+        report="$report_dir/$platform"
+        echo "[$platform] Generating report file: $report"
+        mkdir -p "$(dirname "$report")"
+
+        # Create a report file for the platform.
+        echo "[$platform] Determining software versions..."
+        docker-run "$platform" bash -c '
+            PS4="\$ "
+            set -x
+
+            nextstrain --version
+            nextalign --version
+            nextclade --version
+            augur --version
+            auspice --version
+            python3 -c "from importlib.metadata import version; print(version(\"evofr\"))"
+            datasets --version
+            dataformat version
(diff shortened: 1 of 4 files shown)


devel/validate-platforms
```diff
@@ -40,20 +40,41 @@ main() {
...
             nextclade --version
-            augur --version
             auspice --version
-            python3 -c "from importlib.metadata import version; print(version(\"evofr\"))"
-            datasets --version
-            dataformat version
+            # datasets --version
+            # dataformat version
+
+            pip list 2>&1 | grep -E "'"$tracked_packages_delim"'"
```
victorlin, line 77:
```suggestion
            pip freeze 2>&1 | grep -E "'"$tracked_packages_delim"'"
```

I'd use `pip freeze` over `pip list` since it's a bit more verbose. See the difference in outputs:

`pip list`:

```diff
--- linux/amd64	2023-03-07 10:06:11.000000000 -0800
+++ linux/arm64	2023-03-07 10:06:15.000000000 -0800
@@ -9,6 +9,6 @@
 bcbio-gff                0.6.9
-biopython                1.81
+biopython                1.80
 cycler                   0.11.0
 evofr                    0.1.16
-nextstrain-augur         21.0.0      /nextstrain/augur
+nextstrain-augur         21.0.1      /nextstrain/augur
 nextstrain-cli           6.1.0.post1
```

`pip freeze`:

```diff
--- linux/amd64	2023-03-07 10:18:41.000000000 -0800
+++ linux/arm64	2023-03-07 10:18:43.000000000 -0800
@@ -9,6 +9,6 @@
 bcbio-gff==0.6.9
-biopython==1.81
+biopython==1.80
 cycler==0.11.0
 evofr==0.1.16
-# Editable install with no version control (nextstrain-augur==21.0.0)
+# Editable install with no version control (nextstrain-augur==21.0.1)
 nextstrain-cli==6.1.0.post1
```

devel/validate-platforms
```diff
@@ -40,20 +40,41 @@ main() {
...
+            pandas
+            phylo-treetime
+            pyfastx
+            pysam
+            rethinkdb
+            scikit-learn
+            scipy
+            snakemake
+            xopen
+        )
```
victorlin, line 61: There's some duplication in this script now (it checks versions of Augur and CLI twice). If adding this change, I would factor the code a bit more to track non-pip and pip-installed packages separately.

devel/validate-platforms
corneliusroemer: I'm not going to argue over output but to me this looks like a reformatting without any extra information at all - so not like `-vv` adding `info` and `debug` levels, but just adding extra words to `warn` and `error` 🙃 

It seems to be a 1:1 mapping but with less readable formatting

devel/validate-platforms
corneliusroemer: I'm not sure if `pip list` just shows `pip` installed packages. I think it shows all Python packages. Even if `mamba` installed. But we may not mamba install so we may well have pip-installed them.

Agree that it could be refactored. Just wanted to throw the very useful pip list output in the ring. I thought I'd written an issue to ask for `pip list` output somewhere, maybe `conda-base`, a while ago.

victorlin: > I don't think so 🙃 have a look at the full diff

Oops, I don't know how I skipped over that. Disregard!

devel/validate-platforms
```diff
@@ -40,20 +40,41 @@ main() {
...
-            nextstrain --version
             nextalign --version
             nextclade --version
-            augur --version
             auspice --version
-            python3 -c "from importlib.metadata import version; print(version(\"evofr\"))"
-            datasets --version
-            dataformat version
+            # datasets --version
+            # dataformat version
```
tsibley, line 75: Why commented out?

devel/validate-platforms
```diff
@@ -40,20 +40,41 @@ main() {
...
+            pyfastx
+            pysam
+            rethinkdb
+            scikit-learn
+            scipy
+            snakemake
+            xopen
+        )
+        tracked_packages_string=${tracked_packages[*]}
+        tracked_packages_delim=${tracked_packages_string// /|}
```
tsibley, line 63: Instead of this manipulation plus the `grep -E "'"$tracked_packages_delim"'"` construct, you can hold grep differently:

```
… | grep -Ff <(printf '%s\n' "${tracked_packages[@]}")
```

tsibley:
I really don't think it's reasonable to break the build if one of these selected third-party packages varies in version.

devel/validate-platforms
corneliusroemer: No clue, good point!

corneliusroemer:
> I really don't think it's reasonable to break the build if one of these selected third-party packages varies in version.

Aside from treetime not being third-party (which could be addressed with `treetime --version`, doesn't require `pip list` addition), it would be good to _know_ if there's a difference in version for the major dependencies. I agree erroring is probably too much, but warning would be good. Maybe we could open an issue automatically, or send a Slack message when there's dependency mismatch, so it can be looked into.

tsibley:
> …but warning would be good. Maybe we could open an issue automatically, or send a Slack message when there's dependency mismatch, so it can be looked into.

While we'd probably catch some real issues earlier, I think it the vast majority of the time warnings/notifications/etc would be more often noise than signal and more often busywork than useful tasks. There's a balance between locking everything down super tight and letting stuff float, and I think trying to lock all versions between the two images is tipping too far in one direction.

> Most important from my perspective is to generate the information about what's inside at build time so it can be looked up quickly.

Nod. I guess I see it as easy enough to inspect the images afterwards, and that approach never has any issue of "oops, didn't think to include that version in the report".

victorlin:
Closing since this is out of sync with the target branch, and the motivation is captured by a new issue #141.

PR #136, opened 2023-03-13
chore(deps): update Snakemake to 7.24.1

### LINKED ISSUE #135: Epic: Work towards using more recent Snakemake version
### Context

We currently pin Snakemake to version 5.10.0, which is more than 3 years old.

I think the reasoning was that after that version, a bug/change was introduced that made certain workflows of ours not work with versions higher than 5.10.0

I usually use latest Snakemake in my work and don't encounter issues. So either this bug has been resolved now, or the way we use Snakemake is arcane and no longer supported by current Snakemake.

While pinning Snakemake may not be a bad idea, given Snakemake's tendency for unexpected breaking changes, this version should be updated regularly to something not too old.

I've seen discussions where we couldn't use some new Snakemake feature because we were using 5.10.0. I'm sure there are other reasons not to use an old version.

### Description

It may well be that if we were to start using a more recent version like 7.20, some workflows using the latest docker builds break.

So we should test our major workflows before making a change to Snakemake version so we see how much would have to be changed.

For no-longer updated workflows, they can just use old docker builds - that's the beauty of docker.

Any thoughts @joverlee521 @tsibley @huddlej @victorlin @j23414?

corneliusroemer:
Resolves #135

Should not be merged, but used for testing on our maintained workflows and pipelines that use docker-base

### Testing

- [ ] ncov
- [ ] monkeypox
- [ ] seasonal-flu
- [ ] ncov-ingest
- [ ] ... please add anything I missed

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -184,10 +184,11 @@ RUN pip3 install envdir==1.0.1
 RUN pip3 install awscli==1.18.195
 
 # Install Snakemake and related optional dependencies.
-RUN pip3 install snakemake==5.10.0
+# Pinned to 7.24.1 for stability (2023-03-13)
+RUN pip3 install snakemake==7.24.1
 # Google Cloud Storage package is required for Snakemake to fetch remote files
 # from Google Storage URIs.
-RUN pip3 install google-cloud-storage==2.1.0
+RUN pip3 install google-cloud-storage==2.7.0
 
 # Install epiweeks (for ncov)
 RUN pip3 install epiweeks==2.1.2


Dockerfile
```diff
@@ -184,10 +184,11 @@ RUN pip3 install envdir==1.0.1
 RUN pip3 install awscli==1.18.195
 
 # Install Snakemake and related optional dependencies.
-RUN pip3 install snakemake==5.10.0
+# Pinned to 7.24.1 for stability (2023-03-13)
+RUN pip3 install snakemake==7.24.1
 # Google Cloud Storage package is required for Snakemake to fetch remote files
 # from Google Storage URIs.
-RUN pip3 install google-cloud-storage==2.1.0
+RUN pip3 install google-cloud-storage==2.7.0
```
victorlin, line 191: What's the reason for bumping `google-cloud-storage`?

corneliusroemer: Not sure, it's an optional dependency of snakemake but doesn't seem to be pinned. So could also leave as is.

PR #139, opened 2023-03-15
Add TreeTime

### LINKED ISSUE #134: Add treetime as direct pip3 install?
### Context
I've tried to get `treetime` included in the list of tools that should have consistent versions across docker builds.

The reasoning is that `treetime` is a core Nextstrain component that we should always use the latest version of. Docker builds should be redone when there's a new treetime version.

Right now, treetime gets installed via Augur. I was wondering whether that is sufficient or not - and whether there would be benefit in having it as a separate top level pip install.

victorlin:

Even though TreeTime is implicitly installed in the builder stage as a dependency of Augur, it's more proper to install it explicitly if downstream workflows/analyses will be using it directly.

Also, copy the CLI executable to the final image to enable downstream workflows/analyses to use it directly.

### Related issue(s)

Resolves #134

### Testing

- [x] Checks pass

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -232,6 +232,9 @@ WORKDIR /nextstrain/fauna
 RUN /builder-scripts/download-repo https://github.com/nextstrain/fauna master . \
  && pip3 install --requirement=requirements.txt
 
+# Add Treetime
+RUN pip3 install phylo-treetime
+
 # Augur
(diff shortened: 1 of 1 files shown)


tsibley (approved):

tsibley:
Screenshot from 2023-03-20 12-58-44

Issue #42, opened 2022-05-06
Sporadic error with IQ-TREE v2.1.2


joverlee521:
Documenting an error that I ran into for the first time in the seasonal flu builds that has been reported to IQ-TREE in multiple issues:

- https://github.com/Cibiv/IQ-TREE/issues/175
- https://github.com/iqtree/iqtree2/issues/11
- https://github.com/Cibiv/IQ-TREE/issues/196
- https://github.com/iqtree/iqtree2/issues/47

This seems to be a sporadic error and should be fixed in [v2.2.0.3](https://github.com/iqtree/iqtree2/releases/tag/v2.2.0.3)

danrlu:
Also just saw this in our SARS-CoV-2 build (out of a couple thousands of runs). Thanks for the compilation of issues!

PR #145, opened 2023-05-04
Update pinned software to latest

corneliusroemer:
- Update mafft to v7.520
- Update IQtree to v2.2.2.4
- Update csvtk to v0.25.0
- Update seqkit to v2.4.0
- Update gofasta to v1.2.0
- Update minimamp2 to v2.26
- Update arm jaxlib to v0.4.6
- Update awscli to v1.27.126
- Update snakemake to v7.25.2
- Update google-cloud-storage to v2.8.0
- Update epiweeks to v2.2.0


tsibley:
Is there something motivating these updates at this time beyond "because they're possible"?

If not, we should talk more broadly about why/when to make updates like this. Updates for their own sake are not always productive and often counter-productive.

corneliusroemer:
@tsibley very happy to talk more broadly about pros/cons of updates and how we'd want to do them, on which schedule etc

I noticed that many pinned tools were quite out of date, so I thought I'll go ahead and see what the most up to date versions are to start a discussion, partially motivated by our recent snakemake update experience.

More regular updates could help us reduce the changes necessary for updates if they become necessary as we'll notice any required changes earlier (once we know we need to change something, new code that is up to date won't need to be changed later) [most code that had to be changed for snakemake upgrade was actually written well after we could have been aware that the code we were writing wouldn't be v7 compatible]

We definitely don't need to update every month, but bringing things up to date every half year or year could be good.

Also, this PR is a good motivation to improve how we test images for compatibility, see #147 (thanks for your work there in #148!)

So, absolutely no rush here, but something that'd be good to tackle.

victorlin:
I think updates are good when there is a specific bugfix or improvement in a newer version that benefits users. Otherwise, they're neutral at best, or even detrimental given potential to introduce newer bugs/incompatibilities just by nature of change.

For this PR, I'd approve 96ac3e839a23b645851a28ade13025a51437050b as-is and 3a480478269e1166f3f73090643202fbbd35a160 with added detail that it fixes #42.

PR #146, opened 2023-05-04
Install pango_aliasor

corneliusroemer:
Package maintained by myself to simplify aliasing and dealising of pango lineages

Package is pure Python without dependencies and minimal size

Source code: https://github.com/corneliusroemer/pango_aliasor

See further discussion: https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1683137801980639

Blocking for https://github.com/nextstrain/forecasts-ncov/pull/39

TODO:
- [x] Parallel addition to nextstrain/conda-base, see https://github.com/nextstrain/conda-base/pull/25

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -202,6 +202,8 @@ RUN pip3 install git+https://github.com/cov-lineages/constellations.git@v0.1.1
 RUN pip3 install git+https://github.com/cov-lineages/pango-designation.git@19d9a537b9
 RUN pip3 install pysam==0.19.1
 
+# Install pango_aliasor (for forecasts-ncov)
+RUN pip3 install pango_aliasor==0.3.0
 
 # 4. Add unpinned programs
 


corneliusroemer:
Anything holding this up? It's blocking for https://github.com/nextstrain/forecasts-ncov/pull/39

trvrb:
Looks good to me, though I haven't done any updating to `docker-base` in a while and so I don't know if there's anything we need to do to trigger image rebuild or if this will work automatically.

@tsibley or @victorlin: Okay to merge this?

victorlin (approved):
Looks good to me, though we had bigger picture discussions in Slack, i.e. maybe forecasts-ncov should maintain its own Docker image that extends `nextstrain/base` with `evofr` and `pango_aliasor` instead of relying on them here.

tsibley:
This seems fine, but @corneliusroemer please do follow up sooner than later with the "Parallel addition to nextstrain/conda-base" todo.

tsibley:
@trvrb 

> I don't know if there's anything we need to do to trigger image rebuild or if this will work automatically.

It's fully automatic. …but we're [currently experiencing a CI breakage](https://github.com/nextstrain/docker-base/pull/148#issuecomment-1539203279) due to a change from this morning. A proposed fix is currently being tested. I plan to resolve it one way or another tomorrow morning.

corneliusroemer:
If we were to make more workflow specific images on top of docker-base, we could remove pango_aliasor again. All repos that were to use that package would then just have to make their own images - but that number isn't going to be much larger than 1 anyways. So I don't think that's blocking.

joverlee521:
Merging now that our CI and cache issues have been fixed.

Issue #147, opened 2023-05-05
Run pathogen-repo-ci in CI with new docker-base images produced in CI


corneliusroemer:
### Context

Updating the Snakemake version in docker-base from v5 to v7 was quite involved. One major hurdle was that it wasn't straight forward to test workflows for compatibility with the new image. It would be good if we had more automation helping us check new images don't break anything - or be able to notice issues quickly so that they can be addressed.

### Description

When updating the docker-base image, it would be nice if we could see whether the pathogen CI for various repos still passes with the new docker-base image.

We could manually run these tests in each repo with the new images, but that is tedious. Running pathogen CI as part of docker-base CI may be helpful in quickly spotting issues.

We would probably use the results only in informative ways, not to prevent a new image from being published as pathogen CI can fail for reasons other than the image update.

PR #148, opened 2023-05-05
ci: Test pathogen repo CI builds with the final image

tsibley:
A useful check for if new images will break our pathogen builds.

I included all pathogen repos that already use our pathogen-repo-ci reusable workflow.  It should be minimal effort to maintain this list over time—I expect it to only grow—but perhaps in the future we will want to abstract it out into a shared list of known pathogen repos.

I don't like that we have to copy the build-args for a few of the repos here since it'll be easy for this copy to diverge from the repo's authoritative build-args, but it's necessary for now.  Over time as we work towards increased automation of pathogen builds, I think we can get rid of this build-args copy by further standardizing how each repo configures itself for automation.  For example, instead of specifying build-args in a repo's CI workflow, the args for CI could be stored in a broader workflow metadata file (e.g. nextstrain-workflow.yaml) read by pathogen-repo-ci, or defined by some other convention.

An alternative to directly running pathogen-repo-ci against each repo here would be instead triggering the CI workflows themselves within each repo.  The downside to that is it would divorce the outcomes of those workflows from this one and render them not visible from PRs in this repo.  It would also require updates to each repo to support triggering and passing in of additional parameters (i.e. for the image).  And finally those CI workflows sometimes run other jobs, like linting and other integration tests (e.g. with Cram), that aren't always necessary to run with a new image.

Resolves <https://github.com/nextstrain/docker-base/issues/147>.

### Testing

- [x] Checks pass

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,6 +82,37 @@ jobs:
       env:
         DOCKER_DEFAULT_PLATFORM: ${{ matrix.platform }}
 
+  # Run pathogen repo CI builds with the final image
+  test-pathogen-repo-ci:
+    needs: build
+    strategy:
+      # XXX TODO: Test on multiple platforms via the matrix too, as above?
+      matrix:
+        include:
+          - { pathogen: avian-flu,       build-args: auspice/flu_avian_h5n1_ha.json }
+          - { pathogen: ebola }
+          - { pathogen: lassa }
+          - { pathogen: monkeypox }
+          - { pathogen: mumps }
+          - { pathogen: ncov,            build-args: all_regions -j 2 --profile nextstrain_profiles/nextstrain-ci }
+          - { pathogen: seasonal-flu,    build-args: --configfile profiles/ci/builds.yaml -p }
+          - { pathogen: zika }
+    name: test-pathogen-repo-ci (${{ matrix.pathogen }})
+    uses: nextstrain/.github/.github/workflows/pathogen-repo-ci.yaml@master
+    with:
+      repo: nextstrain/${{ matrix.pathogen }}
+      build-args: ${{ matrix.build-args }}
+      env: |
+        NEXTSTRAIN_DOCKER_IMAGE: ghcr.io/nextstrain/base:${{ needs.build.outputs.tag }}
+
+    # XXX FIXME: Jobs which call reusable workflows (uses: …) can't specify
+    # continue-on-error.¹  That's unfortunate, as we don't want to fail the
+    # workflow because of these.  They're more smoke tests than gating tests.
+    #   -trs, 5 May 2023
+    #
+    # ¹ <https://docs.github.com/en/actions/using-workflows/reusing-workflows#supported-keywords-for-jobs-that-call-a-reusable-workflow>
+    #continue-on-error: true
+
   validate-platforms:
     name: Validate platforms
     needs: build


tsibley:
Ah, it's as I feared and jobs which call reusable workflows can't use the `continue-on-error` job key. This key was not included in [the list of accepted fields](https://docs.github.com/en/actions/using-workflows/reusing-workflows#supported-keywords-for-jobs-that-call-a-reusable-workflow), but I thought it was worth trying anyway as it might be an oversight.

I'll figure out how to rework this to have ~equivalent behaviour without that key, but first I'm going to repush without it to see if the rest of the changes work.

tsibley:
Oh, hmm. To pull from ghcr.io the job needs to be authenticated. That'll require threading some changes thru the shared pathogen-repo-ci workflow. Alternatively, we can switch to the docker.io registry, but then the new test jobs need to move after the push-{build,branch} jobs, and that's potentially annoying for workflow DAG (conditional execution) reasons. And in any case, the pulls from docker.io should ideally still use authentication so we don't hit the low rate limits.

tsibley:
It [works](https://github.com/nextstrain/docker-base/actions/runs/4897729916?pr=148)!

image

Before merge:

- [ ] Wait for merge of https://github.com/nextstrain/.github/pull/38
- [ ] Drop `tmp!` commit ([4471bfc](https://github.com/nextstrain/docker-base/pull/148/commits/4471bfc39dd57fcc0baf858e85916037d86c5dc5))

.github/workflows/ci.yml
```diff
@@ -82,6 +82,39 @@ jobs:
       env:
         DOCKER_DEFAULT_PLATFORM: ${{ matrix.platform }}
 
+  # Run pathogen repo CI builds with the final image
+  test-pathogen-repo-ci:
+    needs: build
+    strategy:
+      # XXX TODO: Test on multiple platforms via the matrix too, as above?
```
corneliusroemer, line 89: Yeah that sounds reasonable!

.github/workflows/ci.yml
```diff
@@ -82,6 +82,39 @@ jobs:
...
+      matrix:
+        include:
+          - { pathogen: avian-flu,       build-args: auspice/flu_avian_h5n1_ha.json }
+          - { pathogen: ebola }
+          - { pathogen: lassa }
+          - { pathogen: monkeypox }
+          - { pathogen: mumps }
+          - { pathogen: ncov,            build-args: all_regions -j 2 --profile nextstrain_profiles/nextstrain-ci }
```
corneliusroemer, line 97: Really nice config with the `build-args`, brilliant!

corneliusroemer (approved):
This is amazing, thanks so much!

corneliusroemer:
Would it make sense to also validate new conda-base environments like this? This could potentially be simple as all we need to do is tell the cli to install and use `--conda` instead of `--docker`.

tsibley:
> Would it make sense to also validate new conda-base environments like this? This could potentially be simple as all we need to do is tell the cli to install and use `--conda` instead of `--docker`.

Yep. I'd probably do this with two additional supporting changes in [`pathogen-repo-ci.yaml` shared workflow](https://github.com/nextstrain/.github/blob/master/.github/workflows/pathogen-repo-ci.yaml):

1. Set up the Conda runtime and run the pathogen build in it _in addition_ to the current use of the Docker runtime. This enables testing in both runtimes for normal pathogen CI (i.e. triggered by changes to the pathogen repos) and also allows the conda-base CI to have a very similar setup as docker-base does with this PR.

2. Add workflow inputs to explicitly disable/enable the runtimes tested. The idea is to default both Docker and Conda runtimes to enabled so normal pathogen CI tests in both, while the uses in docker-base CI and conda-base CI would disable every runtime but themselves so as not to waste cycles / potentially raise false alarms.

.github/workflows/ci.yml
```diff
@@ -82,6 +82,39 @@ jobs:
...
+      repo: nextstrain/${{ matrix.pathogen }}
+      build-args: ${{ matrix.build-args }}
+      env: |
+        NEXTSTRAIN_DOCKER_IMAGE: ghcr.io/nextstrain/base:${{ needs.build.outputs.tag }}
+
+    # XXX FIXME: Jobs which call reusable workflows (uses: …) can't specify
+    # continue-on-error.¹  That's unfortunate, as we don't want to block
+    # subsequent jobs or fail the workflow because of these.  They're more
+    # smoke tests than gating tests.  Would be good to figure out how to adjust
+    # subsequent job conditions (if: …) to get ~equivalent behaviour.
```
joverlee521, line 112: If the pathogen-repo-ci tests shouldn't block the remaining jobs, can it just be forced to be the last step of the workflow?

tsibley: Ah, the comment is maybe a bit misleading because I forgot to update it to remove the "block subsequent jobs" bit. No subsequent jobs actually depend on (`needs: …`) the `test-pathogen-repo-ci` jobs, so failures of the latter won't block any jobs. But failures will still fail the workflow as a whole, which seems possibly undesirable.

tsibley: [Reworded](https://github.com/nextstrain/docker-base/compare/f351e4c8aab40acdd1878ba5bd51586d49b9409b..12000a20d9df02b59cce4aa7f466833fb54df362)

joverlee521: Ah, I see. I think the workflow failing as a whole is a good "notification" of these pathogen-repo-ci failures or else we might miss them. 

Although, thinking aloud on maybe a way around the whole workflow failure:
Add `continue-on-error: ${{ inputs.continue-on-error }}` to the reusable pathogen-repo-ci workflow's `build` job. Then workflows that use the pathogen-repo-ci workflow can pass the input to force the `continue-on-error` behavior.

tsibley: Ahh, yes, that might work!

tsibley: I agree ok to leave them as failing the workflow for now. If we find they're too noisy/uninformative after some time, we can revisit making them smoke-tests only.

tsibley:
This [passed CI for the PR](https://github.com/nextstrain/docker-base/actions/runs/4919734946/jobs/8788422344) but then [failed CI after the merge to master](https://github.com/nextstrain/docker-base/actions/runs/4919742787/jobs/8789546140). The reason why comes down to the difference in `NEXTSTRAIN_DOCKER_IMAGE` values between the two runs:

    NEXTSTRAIN_DOCKER_IMAGE=ghcr.io/nextstrain/base:branch-trs-test-pathogen-repo-ci
    NEXTSTRAIN_DOCKER_IMAGE=ghcr.io/nextstrain/base:build-20230508T211841Z

The `build-*` tag triggers a different code path within Nextstrain CLI's Docker runner for `nextstrain update`:

https://github.com/nextstrain/cli/blob/babc215092465758bf3df88ccf3eaae54add47c3/nextstrain/cli/runner/docker.py#L434-L440

and it ends up trying to enumerate the `build-*` tags of the ghcr.io image by making requests to the docker.io registry. This results in a 400 Bad Request from the latter. Oops.

.github/workflows/ci.yml
tsibley: > Ah, I see. I think the workflow failing as a whole is a good "notification" of these pathogen-repo-ci failures or else we might miss them.

[I implemented](https://github.com/nextstrain/.github/pull/40) the `continue-on-error` pass thru, and the UI that GitHub presents makes it [pretty hard to miss the failures](https://github.com/nextstrain/.github/pull/40#issuecomment-1540545735). So that's nice!

tsibley:
https://github.com/nextstrain/cli/issues/279

tsibley:
#928 should fix the broken CI by ignoring errors in the `test-pathogen-repo-ci` jobs. I can then address the failing jobs separately.

victorlin:
> #928 should fix the broken CI by ignoring errors in the `test-pathogen-repo-ci` jobs

I assume you meant https://github.com/nextstrain/.github/pull/40 + https://github.com/nextstrain/docker-base/pull/150 ?

tsibley:
Indeed. … I don't know where #928 came from!

tsibley:
I [wrote](https://github.com/nextstrain/docker-base/pull/148#issuecomment-1538893769):

> Yep. I'd probably [implement the same testing for the Conda runtime] with two additional supporting changes…

Both supporting changes done in https://github.com/nextstrain/.github/pull/42.

Corresponding updates for docker-base in https://github.com/nextstrain/docker-base/pull/154.

Testing for conda-base in https://github.com/nextstrain/conda-base/pull/27.

PR #150, opened 2023-05-09
ci: Ignore failures of pathogen-repo-ci builds for now

tsibley:
This makes them advisory-only, as they're currently broken.¹

We may also opt to leave them advisory-only longer term, as they might
be noisy and not fit for stopping an image release.

Idea for the pass thru this uses from @joverlee521.²

Related-to: <https://github.com/nextstrain/docker-base/pull/148>

¹ <https://github.com/nextstrain/docker-base/pull/148#issuecomment-1539203279>
² <https://github.com/nextstrain/docker-base/pull/148#discussion_r1187794955>

### Testing

- [x] Checks pass

### Before merge

- [x] Await merge of https://github.com/nextstrain/.github/pull/40
- [x] Drop `tmp!` commit


tsibley:
Merging to fix broken CI.

PR #151, opened 2023-05-09
ci: Test pathogen repo CI builds using images on docker.io instead of ghcr.io

tsibley:
This should fix the currently-broken test-pathogen-repo-ci jobs when our CI runs on master² (and they should continue to work on branches too). It requires a little rearranging of jobs in the workflow, with a little additional and unfortunate complexity due to conditionals.

This works around a Nextstrain CLI bug with registries other than docker.io¹ during `nextstrain update docker` (and `nextstrain setup docker`), which is run as part of our setup-nextstrain-cli action used by this workflow.  Ideally we'll fix that bug and then be able to revert this, especially since we might actually want to condition the pushing to docker.io on the outcome of these test jobs in the future.

¹ <https://github.com/nextstrain/cli/issues/279>
² <https://github.com/nextstrain/docker-base/pull/148#issuecomment-1539203279>

### Testing

- [x] Checks pass

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,30 +82,6 @@ jobs:
       env:
         DOCKER_DEFAULT_PLATFORM: ${{ matrix.platform }}
 
-  # Run pathogen repo CI builds with the final image
-  test-pathogen-repo-ci:
-    needs: build
-    strategy:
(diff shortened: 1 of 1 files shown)


victorlin (commented):
Description and changes make sense 👍

PR #153, opened 2023-05-10
Don't delete the GHCR package

### LINKED ISSUE #131: CI: Transient 403 Forbidden errors when pushing to ghcr.io
While running the CI workflow multiple times over the past few days, I noticed several instances of 403 errors which happen after the build finishes and the image is being pushed to GitHub Container Registry (GHCR). A few examples ([1](https://github.com/nextstrain/docker-base/actions/runs/4298745862/jobs/7493216187), [2](https://github.com/nextstrain/docker-base/actions/runs/4298745862/jobs/7510704448)):

```
ERROR: failed to push ghcr.io/nextstrain/base-native-builder:branch-98-merge: unexpected status: 403 Forbidden
ERROR: failed to push ghcr.io/nextstrain/base-native-builder:branch-98-merge: failed commit on ref "manifest-sha256:40dee82b1e0a00c8568931433a053fa7049ff66fef862a560ebb23b89a2fdd98": unexpected status: 403 Forbidden
```

These are obviously transient because re-runs are successful. However, the rate at which it happened is a bit concerning - [this run](https://github.com/nextstrain/docker-base/actions/runs/4298745862) only succeeded on the 3rd retry.

This seems to be a [common issue](https://github.com/docker/build-push-action/issues/463) with many suggested solutions/workarounds. Here are the suggestions I've extracted, with notes on why they don't apply here (or already have been applied):

1. [Delete the image/package](https://github.com/docker/build-push-action/issues/463#issuecomment-924868666).
    - This does not apply to us since as we automatically delete the image (and entire package if necessary) at the end of each run.
2. [Select "Inherit access from source repository" on the package](https://github.com/docker/build-push-action/issues/463#issuecomment-926229761).
    - We can't try this since the package is transient.
3. If using `GITHUB_TOKEN`, [switch to using a PAT](https://github.com/docker/build-push-action/issues/463#issuecomment-939394233).
    - This has [already been applied](https://github.com/nextstrain/docker-base/blob/d59452c4800c09e7f418c80d70379523a7cf84ef/.github/workflows/ci.yml#L35-L36) since I encountered similar transient 403 errors in the past, and it seemed to help at first but probably not with more observations.
4. [Select the **Read and write permissions** option for **Workflow permissions** in the Actions' settings page of the repository](https://github.com/docker/build-push-action/issues/463#issuecomment-1125443655).
    - This is already set.
5. [Verify that the PAT includes the `write:packages` scope](https://github.com/docker/build-push-action/issues/463#issuecomment-1298538631).
    - It does.
6. [Prevent the Docker login action from logging out at the end of the job.](https://github.com/docker/build-push-action/issues/463#issuecomment-924300445)
    - This was tried in #149 and it did not fix the issue.

If this keeps happening, it might be worth modifying [delete-from-ghcr.js](https://github.com/nextstrain/docker-base/blob/e70bbf003cf5ae4cbf51958505b82e1e82bff630/devel/delete-from-ghcr.js) to retain the package so suggestion (2) can be attempted.

### Other occurrences

- [2023-03-21T21:15:21.8301039Z](https://github.com/nextstrain/docker-base/actions/runs/4483603931/jobs/7883087313)

victorlin:

See commit messages.

### Related issue(s)

- (Hopefully) fixes #131

### Testing

We'll need a few runs to see if errors go away, either because of this fix or GitHub coincidentally fixes something on their end.

- [x] Checks pass
- [x] CI runs don't delete the packages

### DIFF
--- a/devel/delete-from-ghcr.js
+++ b/devel/delete-from-ghcr.js
@@ -98,20 +98,13 @@ module.exports = async ({fetch, octokit, tag, token}) => {
 
       if (deleteVersionError.response.data.message == "You cannot delete the last tagged version of a package. You must delete the package instead.") {
 
-        console.log(`Deleting the package ${org}/${packageName} ...`);
-        try {
-          await octokit.request('DELETE /orgs/{org}/packages/{package_type}/{package_name}', {
-            org: org,
-            package_type: 'container',
-            package_name: packageName,
-          });
-          console.log("Done.");
-        } catch (deletePackageError) {
-          console.log(deletePackageError);
-          console.error(`Could not delete ${org}/${packageName}.`);
-          errorEncountered = true;
-          continue;
-        }
+        // The right thing to do would be to delete the package
+        // ${org}/${packageName}. However, this is a potential cause for
+        // transient 403 errors¹ on GitHub Actions, so we'll keep one tagged
+        // version around as a workaround until the underlying issue is fixed.
+        // ¹ https://github.com/nextstrain/docker-base/issues/131
+        console.log(`Not deleting ${org}/${packageName}:${tag} since that requires deleting the package.`);
+
       } else {
         console.error(`Could not delete ${org}/${packageName}:${tag}.`);
         errorEncountered = true;
--- a/devel/delete-from-ghcr.js
+++ b/devel/delete-from-ghcr.js
@@ -1,4 +1,5 @@
-// Delete Nextstrain images with a specific tag from the GitHub Container registry.
+// Delete Nextstrain images associated with a tagged manifest list from the
+// GitHub Container registry.
 //
 // This file is intended for use by the GitHub action "actions/github-script".
 // To use locally, set `octokit`¹ as an Octokit instance with a personal access
@@ -6,7 +7,7 @@
 // ¹ https://github.com/octokit/core.js#authentication
 // ² https://github.com/settings/tokens/new?scopes=delete:packages
 
-module.exports = async ({octokit, tag}) => {
+module.exports = async ({fetch, octokit, tag, token}) => {
   org = 'nextstrain';
   packages = ['base', 'base-builder'];
 
@@ -14,6 +15,8 @@ module.exports = async ({octokit, tag}) => {
   let errorEncountered = false;
 
   for (const packageName of packages) {
+    // First, get a list of package "version" objects containing information
+    // used for deletion.
     let packageVersions;
     try {
       packageVersions = (await octokit.request('GET /orgs/{org}/packages/{package_type}/{package_name}/versions', {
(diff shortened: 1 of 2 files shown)


victorlin:
I haven't noticed any new occurrences of #131 since this PR forced the packages to be persistent, so I'm hopeful that this is a fix.

devel/delete-from-ghcr.js
```diff
@@ -98,20 +98,13 @@ module.exports = async ({fetch, octokit, tag, token}) => {
...
-          errorEncountered = true;
-          continue;
-        }
+        // The right thing to do would be to delete the package
+        // ${org}/${packageName}. However, this is a potential cause for
+        // transient 403 errors¹ on GitHub Actions, so we'll keep one tagged
+        // version around as a workaround until the underlying issue is fixed.
+        // ¹ https://github.com/nextstrain/docker-base/issues/131
+        console.log(`Not deleting ${org}/${packageName}:${tag} since that requires deleting the package.`);
+
```
tsibley, line 107: _non-blocking_

Hm, kinda hacky, huh?

This makes me want to reconsider using a different registry for CI. Ideally we'd be able to use something like a workflow-local `registry` service, similar to what we do in local dev. GitHub Actions [service containers](https://docs.github.com/en/actions/using-containerized-services/about-service-containers) are job-scoped though not workflow-scoped, so we can't just use those as-is. We could, however, if we configured shared temporary storage (e.g. on S3). Alternatively, we could use some other private container registry for CI, like ECR.

Either way, such a change would also be motivated by the expectation of eventually wanting to distribute our released images via ghcr.io in addition to or instead of docker.io (due to the latter's usage limits, ToS, separate users/auth, etc).

tsibley (approved):

devel/delete-from-ghcr.js
```diff
@@ -28,6 +31,45 @@ module.exports = async ({octokit, tag}) => {
...
+    // This works when a GitHub token with the right permissions is passed in
+    // base64 form as a Bearer token¹.
+    // ¹ https://github.com/orgs/community/discussions/26279#discussioncomment-3251172
+    const res = await fetch(`https://ghcr.io/v2/${org}/${packageName}/manifests/${tag}`,
+      {
+        headers: {
+          Accept: "application/vnd.docker.distribution.manifest.list.v2+json",
+          Authorization: `Bearer ${btoa(token)}`
+        }
+      });
```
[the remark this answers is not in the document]
victorlin: I tried `skopeo`, and don't think it works. See #163.

devel/delete-from-ghcr.js
victorlin: I rewrote this into an issue: #164

PR #155, opened 2023-05-12
Revert "build: Use different cache and output registries"

tsibley:
This reverts commit 917a04806cffdb76dda815b79479f05b1879e31a and fixes the comment added by efbbfb99c8118ccf0d201aa863c88c8a7bbb4d8d.

We need to cache from ghcr.io because otherwise the build of nextstrain/base doesn't find the layers from the just-built nextstrain/base-builder pushed to ghcr.io.  Instead, it seemingly uses layers from an older nextstrain/base-builder on docker.io.

My understanding is that we were always pulling caches from docker.io (_and_ ghcr.io) because docker.io is the default registry when none is specified, and we use --cache-from lines without a registry specified.

See <https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1683932186614689>.

### Testing

- [x] Checks pass

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,8 +11,8 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  # Build multi-platform builder and final images with caching from Docker Hub;
-  # push to GitHub Container Registry.
+  # Build multi-platform builder and final images with caching from Docker Hub
+  # and GitHub Container Registry; push to GitHub Container Registry.
   build:
     runs-on: ubuntu-latest
     steps:
@@ -42,7 +42,7 @@ jobs:
         username: nextstrain-bot
         password: ${{ secrets.GH_TOKEN_NEXTSTRAIN_BOT_MANAGE_PACKAGES }}
 
-    - run: ./devel/build -p linux/amd64,linux/arm64 -c docker.io -o ghcr.io -t $TAG
+    - run: ./devel/build -p linux/amd64,linux/arm64 -r ghcr.io -t $TAG
 
     outputs:
       tag: ${{ env.TAG }}
--- a/devel/build
+++ b/devel/build
@@ -3,10 +3,9 @@
 # Builds the nextstrain/base and nextstrain/base-builder images with useful
 # caching and pushes to a registry.
 #
-# By default this tags images using "latest" and pushes to localhost:5000 with
-# same-registry caching, but you can provide a custom tag with -t <tag> and
-# specify different registries with -o <output_registry> and
-# -c <cache_registry>.
+# By default this tags images using "latest" and pushes to localhost:5000, but
+# you can provide a custom tag with -t <tag> and specify a different
+# registry with -r <registry>.
 #
 # Set CACHE_DATE in your environment to force layers after our custom cache
 # point to be re-built. See the ARG CACHE_DATE line in the Dockerfile for more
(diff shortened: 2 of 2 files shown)


tsibley:
This should fix the issue described by the linked Slack thread. To check, do:

```
$ docker run --rm -it nextstrain/base-builder:branch-cache-from-ghcr.io augur --version
22.0.0
$ docker run --rm -it nextstrain/base:branch-cache-from-ghcr.io augur --version
22.0.0
```

Both should be 22.0.0. If base-builder is 22.0.0 and base is 21.1.0, then the bug isn't fixed.

Someone else (maybe @huddlej) should check next week and merge if it looks good.

corneliusroemer (commented):
Tested as explained by @tsibley and it doesn't seem to work for me - or do I need to build first?

```log
$ docker run --rm -it nextstrain/base-builder:branch-cache-from-ghcr.io augur --version
Unable to find image 'nextstrain/base-builder:branch-cache-from-ghcr.io' locally
branch-cache-from-ghcr.io: Pulling from nextstrain/base-builder
Digest: sha256:cb6a55e7c4bd4d9efb9f8fd6ed900d1a11f8f481096c22db794d5b2f3c9347e3
Status: Downloaded newer image for nextstrain/base-builder:branch-cache-from-ghcr.io
augur 22.0.0
$ docker run --rm -it nextstrain/base:branch-cache-from-ghcr.io augur --version
Unable to find image 'nextstrain/base:branch-cache-from-ghcr.io' locally
branch-cache-from-ghcr.io: Pulling from nextstrain/base
Digest: sha256:c50bf9fdca3b52ec43458a7d11188f8778bf12aa06a933de8be6d4cc01ca9833
Status: Downloaded newer image for nextstrain/base:branch-cache-from-ghcr.io
augur 21.1.0
```

huddlej:
Darn, I get the same result as @corneliusroemer. @tsibley and @victorlin are out this week, so we may need to continue troubleshooting on our own, Cornelius...

joverlee521:
Yeah, this isn't resolving the issue. 
Looking at the "Created" date for the images:
```
$ docker inspect -f '{{.Created}}'  nextstrain/base-builder:branch-cache-from-ghcr.io
2023-05-12T23:43:54.823419242Z
$ docker inspect -f '{{.Created}}' nextstrain/base:branch-cache-from-ghcr.io
2023-05-08T22:30:15.085923811Z
```

joverlee521:
Looking at the test run logs for this PR, it seems like the [new branch ghcr.io builder image is being pulled](https://github.com/nextstrain/docker-base/actions/runs/4963854370/jobs/8883454206#step:8:12449) when building the final image. However, the COPY commands from the builder to the final image are [still using the cache](https://github.com/nextstrain/docker-base/actions/runs/4963854370/jobs/8883454206#step:8:12484). 

This leads to my current question of how does Docker handle multiple `--cache-from` sources. I did not find anything in the [docker docs](https://docs.docker.com/engine/reference/commandline/buildx_build/#cache-from), but I did find an old [comment](https://github.com/moby/moby/pull/26839#issuecomment-277383550) on the expected behavior:

> When using multiple --cache-from they are checked for a cache hit in the order that user specified. If one of the images produces a cache hit for a command only that image is used for the rest of the build.

I'm going to try go reorder our [`--cache-from` sources](https://github.com/nextstrain/docker-base/blob/50e442a8bbaafd7b3cbb36c6e703c96c6b42eae7/devel/build#L79-L86) for the final image to see if that will affect the outcome.

joverlee521:
Well, reordering did not work 😞 

Instead of continuing to bang my head on this, I just [removed](https://github.com/nextstrain/docker-base/pull/155/commits/997b702d0d3ad33cfbcd013a533e1e2d5127d207) the `--cache-from` that sources the base-builder image from docker.io to ensure that we only use the latest base-builder image that is on ghcr.io.

joverlee521:
Urgh, so it seems like the base-builder image is not the issue since removing the cache-from docker.io did not fix this issue. 

Somehow when building the final image, Docker is not detecting the changes from the final image cache. The latest push removes the cache for the final image from docker.io so it should run all of the COPY commands.

joverlee521:
I _think_ I found the underlying problem that is leading to the weird cache. 

In a recent [commit](https://github.com/nextstrain/docker-base/pull/133/commits/ba41dea3d13444050c08cfa517f70b14a96f6ba1), the `CACHE_DATE` arg was removed from the final build stage. I think that means when Docker is building the final image, it will not be able to find the base-builder cache layer that uses the same `CACHE_DATE`. If we provide it again, it should be able to figure the correct caches to use for building the final image.

PR #156, opened 2023-05-16
wip: fast build only amd64

corneliusroemer:

QEMU arm builds slow builds down _a lot_. With AMD only, CI would take around 10 min, with ARM as well, it's 60min.

For faster feedback loops it could make sense to be able to run CI with AMD only.

The way I've set it up is definitely not good, but it's a POC. Could be helpful for @joverlee521 while working on #155 for example.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,13 @@
 name: CI
-on: [push, pull_request, workflow_dispatch]
+on:
+  push:
+  pull_request:
+  workflow_dispatch:
+    inputs:
+      platforms:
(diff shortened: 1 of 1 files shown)


tsibley:
Closing since we're not going to use this PR as-is.

We definitely would benefit from a faster CI feedback loop, but I think routinely disabling one half of the build would introduce other bugs. Would be good to do more thinking about how to make platforms an optional switch for some PRs, those where it would specifically be beneficial. One alternative way is temporary commits on the PR branch that edit the platforms used, which are then dropped and retested with full CI before merge.

PR #159, opened 2023-06-06
Tests

tsibley:
See commit messages.

### Testing

- [ ] Checks pass

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 SHELL := bash -euo pipefail
 
-.PHONY: local-image
+.PHONY: local-image test
 
 local-image:
 	./devel/start-localhost-registry || true
@@ -11,3 +11,7 @@ local-image:
 	docker image pull localhost:5000/nextstrain/base:latest
 	
 	./devel/stop-localhost-registry
+
+test:
+	rm -f tests/*.t.err
+	cram -v tests/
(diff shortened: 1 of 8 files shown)


tsibley:
The first few CI failures were due to unanticipated differences between my local testing and CI, and those are now fixed. The latest CI failure is because of #131. However, https://github.com/nextstrain/docker-base/pull/158 is based on the branch in this PR and it succeeded, which indicates this PR would likely also pass CI if not for the transient errors.

Makefile
```diff
@@ -0,0 +1,13 @@
+SHELL := bash -euo pipefail
+
+.PHONY: local-image
+
+local-image:
+	./devel/start-localhost-registry || true
+	./devel/build
```
victorlin, line 7: This uses [default argument values](https://github.com/nextstrain/docker-base/blob/1b0d1998c8cb03cbd0b5eb630fcd86fdfb247c99/devel/build#L16-L19), which includes `platform=linux/amd64`. I develop on Apple silicon and want to use `platform=linux/arm64` for faster builds. This `make` shortcut doesn't provide a way to specify that.

Maybe the default platform in `devel/build` can be inferred from the host machine? That way, this `make` shortcut would build without emulation on both `amd64` and `arm64`.

tsibley: Ahh, yeah, that's a good point. I'll make adjustments.

tsibley: @victorlin see https://github.com/nextstrain/docker-base/pull/159/commits/c2921fe32302583b728b5f592a11112a88568db8

PR #160, opened 2023-06-08
Improve storage for local dev

tsibley:
Fixes for things I ran into during recent local dev sessions. See commit messages for details.

### Testing

- [x] Checks pass (though nothing here is used in CI)

### DIFF
--- a/devel/clean
+++ b/devel/clean
@@ -1,0 +1,61 @@
+#!/bin/bash
+#
+# Cleans up after the build process by removing build artifacts and caches.
+#
+set -euo pipefail
+
+# Set defaults and fixed values.
+repo="$(cd "$(dirname "$0")/.."; pwd)"
+registry=nextstrain-local-registry
+registry_addr=localhost:5000
+build_dir="$repo"/build
+builder=nextstrain-builder
+
+# Read command-line arguments.
+while getopts "r:" opt; do
+    case "$opt" in
+        r) registry_addr="$OPTARG";;
+        *) echo "Usage: $0 [-r <registry>]" 1>&2; exit 1;;
+    esac
+done
+
+
+echo "--> Deleting $registry container"
+
+if docker container inspect "$registry" &>/dev/null; then
+    "$repo"/devel/stop-localhost-registry
+else
+    echo "skipped; $registry container does not exist"
+fi
+
+
+echo "--> Deleting artifacts in $build_dir"
+
+if [[ -d "$build_dir" ]]; then
+    du -hs "$build_dir"
+    rm -rf "$build_dir"
+else
+    echo "skipped; $build_dir does not exist"
+fi
+
+
+echo "--> Deleting $builder (and its caches)"
+
+if docker buildx inspect "$builder" &>/dev/null; then
+    docker buildx du --builder "$builder"
+    docker buildx rm --builder "$builder"
+else
+    echo "skipped; $builder does not exist"
+fi
+
+
+echo "--> Deleting local images pulled from $registry_addr"
+
+for image in "$registry_addr"/nextstrain/base{,-builder}; do
+    for id in $(docker image ls -q "$image"); do
+        docker image rm "$id"
+    done
+done
+if [[ -z "${id:-}" ]]; then
+    echo "skipped; no local images"
+fi
--- a/devel/start-localhost-registry
+++ b/devel/start-localhost-registry
@@ -15,7 +15,7 @@ IMAGE=registry:2
 
 docker run \
     --detach \
-    --publish "$PORT":"$PORT" \
+    --publish 127.0.0.1:"$PORT":"$PORT" \
     --restart always \
     --name "$NAME" \
     "$IMAGE" \
--- a/devel/start-localhost-registry
+++ b/devel/start-localhost-registry
@@ -13,4 +13,10 @@ NAME=nextstrain-local-registry
 # Docker image that provides the registry service.
 IMAGE=registry:2
 
-docker run -d -p "$PORT":"$PORT" --restart always --name "$NAME" "$IMAGE" > /dev/null
+docker run \
+    --detach \
+    --publish "$PORT":"$PORT" \
+    --restart always \
+    --name "$NAME" \
+    "$IMAGE" \
+    > /dev/null
--- a/devel/start-localhost-registry
+++ b/devel/start-localhost-registry
@@ -4,6 +4,11 @@
 #
 set -euo pipefail
 
+# Paths to use for registry data.
+REPO="$(cd "$(dirname "$0")/.."; pwd)"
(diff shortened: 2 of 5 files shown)


devel/clean
```diff
@@ -0,0 +1,51 @@
...
+
+if docker buildx inspect "$builder" &>/dev/null; then
+    docker buildx du --builder "$builder"
+    docker buildx rm --builder "$builder"
+else
+    echo "skipped; $builder does not exist"
+fi
+
+
+echo "--> Deleting local images"
```
victorlin, line 42:
```suggestion
echo "--> Deleting local images pulled from localhost:5000"
```

Clarify that this does not delete local images pulled from elsewhere (e.g. Docker Hub).

devel/start-localhost-registry
```diff
@@ -18,5 +23,7 @@ docker run \
     --publish 127.0.0.1:"$PORT":"$PORT" \
     --restart always \
     --name "$NAME" \
+    --volume "$DATA":/var/lib/registry \
```
victorlin, line 26: _(non-blocking)_

Upon `make clean` I noticed that the new `build/` directory had a size of 2.1G. I tried to find the accumulated volumes from previous registry usage, but none had large sizes except for the builder volume which is used elsewhere. Building from `master` also does not produce any new volumes in `docker volume ls`.

I think I'm missing something here. Without the custom `--volume`, do the transient volumes get pruned somehow?

devel/clean
```diff
@@ -0,0 +1,51 @@
...
+
```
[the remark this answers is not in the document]
tsibley: Hmm. I think you're mistaken. It deletes what's shown in `docker image ls`, not the local registry. (The local registry's copies are deleted with `rm -rf "$build_dir"` earlier.)

`docker image pull localhost:5000/nextstrain/base:latest` results in `localhost:5000/nextstrain/base:latest` being available, e.g. in `docker image ls`. It doesn't result in `nextstrain/base:latest` being available.

devel/clean
```diff
@@ -0,0 +1,51 @@
...
+for image in localhost:5000/nextstrain/base{,-builder}; do
```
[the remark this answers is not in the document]
tsibley: Fair enough. I skipped parameterization here because it seemed unnecessary, but I'll thread it thru.

(The overhead of threading thru params is why I prefer env vars, e.g. `LOCAL_REGISTRY=…`, over command-line arguments when there are many separate programs co-operating over shared resources.)

devel/clean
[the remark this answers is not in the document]
victorlin: Oops, I realized this later and thought I deleted this comment, but looks like that didn't happen. You're right!

devel/clean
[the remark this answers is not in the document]
victorlin: I'd be ok with an env var for the local registry, if you want to make that change here.

devel/start-localhost-registry
tsibley: Hmm. They don't get pruned for me, hence the reason for the change, and I can see them in `docker volume ls`. On `master`, for example:

    $ docker volume ls 
    DRIVER    VOLUME NAME

    $ ./devel/start-localhost-registry 

    $ docker volume ls 
    DRIVER    VOLUME NAME
    local     85e2e9bfedc4a83eea0f00b978133d37b26fbd020294143b40f9c50d32291d87

    $ docker container inspect nextstrain-local-registry -f '{{json .Mounts}}' | jq
    [
      {
        "Type": "volume",
        "Name": "85e2e9bfedc4a83eea0f00b978133d37b26fbd020294143b40f9c50d32291d87",
        "Source": "/var/lib/docker/volumes/85e2e9bfedc4a83eea0f00b978133d37b26fbd020294143b40f9c50d32291d87/_data",
        "Destination": "/var/lib/registry",
        "Driver": "local",
        "Mode": "",
        "RW": true,
        "Propagation": ""
      }
    ]

    $ ./devel/stop-localhost-registry 

    $ docker volume ls 
    DRIVER    VOLUME NAME
    local     85e2e9bfedc4a83eea0f00b978133d37b26fbd020294143b40f9c50d32291d87

    $ docker container inspect nextstrain-local-registry -f '{{json .Mounts}}' | jq
    Error: No such container: nextstrain-local-registry

Perhaps something is auto-pruning them for you, e.g. Docker Desktop or a newer Docker version?

tsibley: What does `docker system df` report for you?

devel/clean
tsibley: Done with repush in 23a8431.

devel/clean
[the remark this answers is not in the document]
tsibley: Parameterized with repush in 23a8431.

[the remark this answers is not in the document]
tsibley: I'll save any switch from command-line args to env vars for another time!

devel/start-localhost-registry
victorlin: Oh nevermind, I see the 2GB volume now. I think Docker Desktop's UI was displaying an incorrect size, or I misread earlier. 

I can reproduce the Docker Desktop UI to show these incorrect sizes after a change in "status":

image

Clicking out then back into the "Volumes" view shows numbers that better reflect what's reported by `docker system df -v`:

```
Local Volumes space usage:

VOLUME NAME                                                        LINKS     SIZE
3647c4b5c1767b95a75be5bcc79ede3ff2cce3e9b52759361f6b3cd3cc0cb96d   0         0B
81f2fbdfd9fa34492dfb49e6761ad50f845c380275cd4a5267276cdc23a39e85   0         2.13GB
buildx_buildkit_nextstrain-builder0_state                          1         5.456GB
d99f38bd772f04ead379fb7eadc20de6589e9c66dd64d3eb3ad605f2c36dd63f   0         0B
f974ecaf61113fe8596ba6bcceaadb364a7435e071a060383fac9e1934030097   0         0B
823334b73184b6465a727efd7835a373a0426650e734a9f4a7c5a85a5ac11f59   0         2.13GB
c0c7d34820a0ebb85914fcbdfe7d6981ff3877ee394866bf6476df506b18a4b1   1         0B
3002016aaf02647af6150de06e83ceec9544bfb8a207a94f801dfe5506764f81   0         0B
fc3b5d46e5200e3ae2ccf1302e1b2c2780a13a17f8cf80880eda6b3519dcc11f   0         0B
```

tsibley: :upside_down_face:

victorlin (approved):
I'm not sure what happened with CI, but changes look good.

tsibley:
(CI failures were a GitHub Actions outage.)

tsibley:
Rebased onto master, including a small fix caught by the new ShellCheck linting.

I'll wait for PR CI to pass before merging.

PR #163, opened 2023-06-21
Use skopeo for cleanup

victorlin:

The new script, devel/delete-images, is much simpler than the previous implementation which called out to the GitHub REST API and Docker Registry API. Those were lower-level requests that involved getting package version objects to map tags/digests to GHCR IDs and using those IDs to delete the objects.

Organization of the script is based on devel/copy-images, which also uses skopeo.

### Related issue(s)

- Motivated by https://github.com/nextstrain/docker-base/pull/153#discussion_r1232909290

### Testing

- [ ] Checks pass


victorlin:
This worked on a local registry, but [failed on CI](https://github.com/nextstrain/docker-base/actions/runs/5337673478/jobs/9674039056#step:4:32):

```
deleting ghcr.io/nextstrain/base:branch-victorlin-use-skopeo-for-cleanup: unsupported
```

Unfortunately, it looks like GitHub Container Registry does not implement that part of the registry API, through similar observations:

- https://github.com/containers/skopeo/issues/1880#issuecomment-1544470232
- https://github.com/orgs/community/discussions/24974#discussioncomment-5875737

`skopeo delete` can't be used on ghcr.io until there is support on the registry side.

PR #165, opened 2023-06-22
Omit dev deps when installing Auspice

tsibley:
This avoids disk-heavy deps like Puppeteer, which bundle a whole Chromium install.  Based on some rough comparisons, I expected this to shave about 600MB from the uncompressed image size and a local test build bore that out.

The node_modules/ tree is infamously large and bloated.  The usual culprits are non-source files that are commonly included in package distributions but not needed at run time.  So there is surely more we could shave off here, but this is a huge easy start.

### Testing

- [x] Checks pass

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -214,7 +214,7 @@ RUN curl -fsSL -o /final/bin/nextclade2 https://github.com/nextstrain/nextclade/
 # ⁴ https://github.com/mapbox/node-pre-gyp/blob/v1.0.10/lib/node-pre-gyp.js#L186
 WORKDIR /nextstrain/auspice
 RUN /builder-scripts/download-repo https://github.com/nextstrain/auspice release . \
- && npm update && npm install && npm link
+ && npm install && npm link
 
 # Add NCBI Datasets command line tools for access to NCBI Datsets Virus Data Packages
 RUN curl -fsSL -o /final/bin/datasets https://ftp.ncbi.nlm.nih.gov/pub/datasets/command-line/v2/linux-${TARGETARCH}/datasets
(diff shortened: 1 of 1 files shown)


tsibley:
I think this is getting skunked by the old version of npm (v6) in the image, which I only noted because I saw this warning float by:

    npm WARN read-shrinkwrap This version of npm is compatible with lockfileVersion@1, but package-lock.json was generated for lockfileVersion@2. I'll try to do my best with it!

victorlin:
> I think this is getting skunked by the old version of npm (v6) in the image

#122, hopefully to be merged soon, will bring a newer version of npm.

tsibley:
Confirmed that with newer npm in the container this works fine.

```patch
commit d38eb85c003fb738cec4a7df7aabc768ec610149
Author: Thomas Sibley <tsibley@fredhutch.org>
Date:   Fri Jun 23 10:11:25 2023 -0700

    wip! upgrade npm

diff --git a/Dockerfile b/Dockerfile
index 569bfd7..e248fdf 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,7 +35,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 # Install a specific Node.js version
 # https://github.com/nodesource/distributions/blob/0d81da75/README.md#installation-instructions
 RUN curl -fsSL https://deb.nodesource.com/setup_14.x | bash - \
- && apt-get update && apt-get install -y nodejs
+ && apt-get update && apt-get install -y nodejs && npm install -g npm@8
 
 # Used for platform-specific instructions
 ARG TARGETPLATFORM
```

Will test some more locally, clean up, and then mark for review.

tsibley:
Dropped the `npm install -g npm@8` bit since #122 is in now and repushed.

victorlin (approved):

PR #170, opened 2023-06-30
Upgrade Node.js to v18

victorlin:

Node.js v16 will reach end-of-life status on 2023-09-11.

Node.js v18 should also come with performance improvements during run time.

### Related issue(s)

- Prompted by https://github.com/nextstrain/docker-base/pull/122#issuecomment-1590159035

### Testing

- [x] Checks pass

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,7 +42,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 
 # Install a specific Node.js version
 # https://github.com/nodesource/distributions/blob/0d81da75/README.md#installation-instructions
-RUN curl -fsSL https://deb.nodesource.com/setup_16.x | bash - \
+RUN curl -fsSL https://deb.nodesource.com/setup_18.x | bash - \
  && apt-get update && apt-get install -y nodejs
 
 # Used for platform-specific instructions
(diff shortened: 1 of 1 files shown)


tsibley (approved):
This seems good to me… though I'm not totally sure how to evaluate what it could potentially break. Auspice should be just fine, and we're only using Node for Auspice, right?

victorlin:
Yes Auspice should be fine. "We" (the Nextstrain core team) are only using Node.js for Auspice. There's nothing stopping other users of the Docker image from using Node.js for other purposes, though the chance of that happening + breakage from this upgrade seems unlikely to me.

If we had some sort of changelog for the Docker image then this would be worth mentioning, but that's not in place.

victorlin:
The PR-triggered CI runs are failing due to what should be unrelated reasons, but I'll still hold off until it passes.

- Attempts [1](https://github.com/nextstrain/docker-base/actions/runs/5425758934/attempts/1), [2](https://github.com/nextstrain/docker-base/actions/runs/5425758934/attempts/2) failed on the same action, which succeeded on attempt 3 so those were transient.
- Attempt [3](https://github.com/nextstrain/docker-base/actions/runs/5425758934/attempts/3) failed because it was re-run with "failed jobs only". That option doesn't work with our CI setup because the successful built images that are pushed to GHCR get deleted upon cleanup of failed runs.

tsibley:
> If we had some sort of changelog for the Docker image then this would be worth mentioning, but that's not in place.

Agreed! A changelog for our runtimes would be useful. In particular because our runtimes intentionally do not offer any compat guarantees between versions. This is for a few reasons, but the primary one is that it would be a _lot_ of maintenance overhead (and difficult) to make those guarantees. The expectation is that if you require strict behavioral compat, you stick with a specific runtime version. We could document this better… that's part of the work I want to do on runtime documentation (https://github.com/nextstrain/cli/issues/288).

> That option doesn't work with our CI setup…

Oh right, I knew this but had forgotten. :man_facepalming:

PR #171, opened 2023-07-06
Update jaxlib for evofr on linux/arm64

joverlee521:
### Description of proposed changes

The latest release of evofr included updates to its dependencies that updated the minimum version for jaxlib.¹ I'm using the latest release of jaxlib available in the user forked repo since there's still no official pre-built binaries for linux/arm64.²

I did not realize this change was needed until I saw the "Validate Platform" job fail in the CI run³ that I triggered to update evofr to v0.1.20.

¹ https://github.com/blab/evofr/commit/46f744b605c258d6f3d3b7b3ac8bc877f8e4ba22 
² https://github.com/google/jax/issues/7097
³ https://github.com/nextstrain/docker-base/actions/runs/5480222698

### Testing

- [ ] Checks pass

Issue #119, opened 2022-12-16
Install jaxlib from a Nextstrain-managed repository


victorlin:
_Originally from https://github.com/nextstrain/docker-base/pull/50#discussion_r1046541339_

On the `linux/arm64` platform, `jaxlib` is installed from a pre-built binary provided by an external GitHub user:

https://github.com/nextstrain/docker-base/blob/960fb5bb8a7abdfbaaf75579019101d0b3067940/Dockerfile#L172-L179

We should look into building and repackaging this library ourselves for security and reliability reasons.

victorlin:
An example of how the unofficial binary is not reliable: https://github.com/nextstrain/docker-base/pull/173#issuecomment-1639058770

victorlin:
I'm looking into building JAX locally.

victorlin:
I have an unfinished attempt at [victorlin/jax](https://github.com/victorlin/jax). Summarizing the progress here since I plan to focus on other things for now.

The goal was to try to compile an AARCH64 version of JAX on a GitHub-hosted Linux runner (AMD64).

I briefly tried building via Docker under emulation, which was [unsuccessful](https://github.com/victorlin/jax/actions/runs/5593165693/job/15149873309). Even if it were successful, it would probably take longer than the maximum allowed GitHub Actions workflow run time.

Then, I focused on cross-compiling since that [has been done before](https://github.com/google/jax/issues/7097#issuecomment-1216826398). I attempted this first with then without Docker (for easier debugging of individual commands as GitHub Actions steps). With a 3.5 hour [run](https://github.com/victorlin/jax/actions/runs/5605615210) time, I was able to build a file `jaxlib-0.4.9-cp310-cp310-manylinux2014_aarch64.whl` (a downloadable artifact of the linked run). However, I'm not sure that this wheel works because `auditwheel repair` failed on the file. I think the failure makes sense because I didn't use a manylinux Docker image. To address that, I [tried](https://github.com/victorlin/jax/compare/8be9738dfd1e25a26d97fde07da364713110d662...a9de12a09e7941d4e77c2c14c999927ab1da8bc4) building with [`messense/manylinux2014-cross:x86_64`](https://github.com/rust-cross/manylinux-cross) but had troubles running the setup commands before the actual JAX build.

victorlin:
[...]

PR #173, opened 2023-07-17
Jax and jaxlib

joverlee521:
Pin jax version for evofr on linux/arm64 since jax has a minimum version requirement for jaxlib that is _not_ checked by package version constraints in pip.¹ Instead, the jaxlib version is checked at runtime. This led to a mismatch of versions when only pinning jaxlib.²

Also updating both the latest available version of jaxlib in the user forked repo since I'm already updating the Dockerfile. 

¹ https://jax.readthedocs.io/en/latest/jep/9419-jax-versioning.html#how-are-jax-and-jaxlib-versioned
² https://github.com/nextstrain/forecasts-ncov/pull/50#issuecomment-1629841748

### Related issue(s)

Follow-up to https://github.com/nextstrain/docker-base/pull/171

### Testing

- [x] Checks pass

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -253,12 +253,16 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 
 # 1. Install programs via pip
 
-# Install jaxlib on linux/arm64
+# Install jaxlib & jax on linux/arm64
 # jaxlib, an evofr dependency, does not have official pre-built binaries for
 # linux/arm64. A GitHub user has provided them in a fork repo.
 # https://github.com/google/jax/issues/7097#issuecomment-1110730040
+# Also hard-coding jax version here since it needs to match the jaxlib version
+# The minimum version requirement for jaxlib is checked at runtime rather than by pip
+# https://jax.readthedocs.io/en/latest/jep/9419-jax-versioning.html#how-are-jax-and-jaxlib-versioned
 RUN if [[ "$TARGETPLATFORM" == linux/arm64 ]]; then \
       pip3 install https://github.com/yoziru/jax/releases/download/jaxlib-v0.4.6/jaxlib-0.4.6-cp310-cp310-manylinux2014_aarch64.manylinux_2_17_aarch64.whl \
+          jax==0.4.6 \
       ; \
     fi
 


joverlee521:
@marlinfiggins is [running into an import error](https://github.com/nextstrain/forecasts-ncov/pull/50#issuecomment-1638952873) with version 0.4.13: 

```
import jax.numpy as jnp
  File "/usr/local/lib/python3.10/site-packages/jax/__init__.py", line 35, in <module>
    from jax import config as _config_module
  File "/usr/local/lib/python3.10/site-packages/jax/config.py", line 17, in <module>
    from jax._src.config import config  # noqa: F401
  File "/usr/local/lib/python3.10/site-packages/jax/_src/config.py", line 24, in <module>
    from jax._src import lib
  File "/usr/local/lib/python3.10/site-packages/jax/_src/lib/__init__.py", line 92, in <module>
    import jaxlib.xla_client as xla_client
  File "/usr/local/lib/python3.10/site-packages/jaxlib/xla_client.py", line 26, in <module>
    from . import xla_extension as _xla
ImportError: /usr/local/lib/python3.10/site-packages/jaxlib/xla_extension.so: undefined symbol: PyObject_Vectorcall
```

I found [`PyObject_Vectorcall` in the CPython docs](https://docs.python.org/3/c-api/call.html#c.PyObject_Vectorcall), so I'm guessing something went wrong during the build? 

I'm going to drop https://github.com/nextstrain/docker-base/pull/173/commits/331b798d5415d623b4a25c806715fe04f19221b2 to see if that resolves the issue.

Dockerfile
```diff
@@ -253,12 +253,16 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
 # 1. Install programs via pip
 
-# Install jaxlib on linux/arm64
+# Install jaxlib & jax on linux/arm64
 # jaxlib, an evofr dependency, does not have official pre-built binaries for
 # linux/arm64. A GitHub user has provided them in a fork repo.
 # https://github.com/google/jax/issues/7097#issuecomment-1110730040
+# Also hard-coding jax version here since it needs to match the jaxlib version
+# The minimum version requirement for jaxlib is checked at runtime rather than by pip
+# https://jax.readthedocs.io/en/latest/jep/9419-jax-versioning.html#how-are-jax-and-jaxlib-versioned
```
victorlin, line 262: It kinda sucks that we have to rely on an unofficial distribution of pre-built binaries plus manage the "peer dependencies" here, all for a dependency of evofr. Sadly this is the price we pay for supporting mutli-arch images.

I'm guessing the `linux/amd64` image variant works fine since the `jax`/`jaxlib` versions are resolved to when running `pip3 install evofr`. However, usage of that image might be noticeably slower on Apple silicon.

victorlin (approved):

victorlin:
> I found [`PyObject_Vectorcall` in the CPython docs](https://docs.python.org/3/c-api/call.html#c.PyObject_Vectorcall), so I'm guessing something went wrong during the build?

This looks like an issue with the unofficial binary. I confirmed that the official 0.4.13 works locally:

```sh
pip install jaxlib==0.4.13 jax==0.4.13
python -c "import jax.numpy"  # this works
```

and the unofficial one doesn't:

```dockerfile
FROM python:3.10-slim-bullseye

RUN pip install https://github.com/yoziru/jax/releases/download/jaxlib-v0.4.13/jaxlib-0.4.13-cp310-cp310-manylinux2014_aarch64.whl jax==0.4.13

RUN python -c "import jax.numpy"  # this fails
```

corneliusroemer:
Are we sure that we should get jax 0.4 instead of 0.3, I faintly remember that there were some issues with 0.4

Checking with evofr dev :) 

@marlinfiggins can you check that we're getting the right versions of everything here?

joverlee521:
> Are we sure that we should get jax 0.4 instead of 0.3, I faintly remember that there were some issues with 0.4

@corneliusroemer The original update to jaxlib 0.4 in #171 was required for evofr v0.1.20 since the [dependencies were updated](https://github.com/blab/evofr/commit/46f744b605c258d6f3d3b7b3ac8bc877f8e4ba22#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711).

tsibley:
> Pin jax version for evofr on linux/arm64 since jax has a minimum version requirement for jaxlib that is _not_ checked by package version constraints in pip.¹ Instead, the jaxlib version is checked at runtime.

The jax/jaxlib dependency situation seems like a mess in more ways than one:

1. Required dependencies which must be satisfied manually
2. No pre-built binaries for non `x86_64` archs
3. Different versions/builds for CPU vs. GPU vs. TPU support

(1) is addressable several ways, including using one of the Python package extras to get an appropriate jaxlib installed. The most recent jax declares these extras, for example, one of which we could use:

```
…

[ci]
jaxlib==0.4.12

[cpu]
jaxlib==0.4.13

…

[minimum-jaxlib]
jaxlib==0.4.11

…
```

(2) is addressable with [a bunch of work](https://github.com/nextstrain/docker-base/issues/119).

(3) is really thorny, though, as addressing it seemingly requires further branching of runtimes (a non-starter for us, I think) assuming we can't actually install jaxlib in a way that supports accelerated compute (GPU, TPU) when available with a fallback to unaccelerated (CPU).

joverlee521:
Merging this as a patch for the current version mismatch for jax/jaxlib since the [test image allowed Marlin to run things locally](https://github.com/nextstrain/forecasts-ncov/pull/50#issuecomment-1642871016).

More robust jax/jaxlib dependency handling can be added later.

PR #176, opened 2023-09-09
Bump snakemake pin to 7.32.3

corneliusroemer:
Since our our last snakemake pin bump ~6 months ago
snakemake has added support for a default profile
that allows reducing required command line flags to
a mininmum. Without default profile, one always either
needs to explicitly specify a profile or set things like
`--cores all` which is very annoying.

Now, you can simply put those flags in `profiles/default/config.yaml`
and they'll automatically be used when you invoke `snakemake`

7.32.3 appears to be reasonably stable, no major bugs reported in a month
I've used it in production for a month now, without issues.

Checklist:

- [ ] Once this is agreed upon, we should mirror the bump to conda-base

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -273,8 +273,8 @@ RUN pip3 install envdir==1.0.1
 RUN pip3 install awscli==1.18.195
 
 # Install Snakemake and related optional dependencies.
-# Pinned to 7.24.1 for stability (2023-03-13)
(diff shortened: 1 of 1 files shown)


victorlin (approved):
Seems reasonable! Thanks for looking into the new features.

corneliusroemer:
Actually, we're already at 7.32.3 in conda-base, and there have been no complaints, so merging

PR #178, opened 2023-09-09
Add nano and vim-tiny for usability in interactive prompt

### LINKED ISSUE #177: Include basic editor(s): nano and vi(m) [tiny or minimal]
### Context
Nextstrain shell is very useful to get a reliable no-worry environment. Sometimes I'd like to quickly edit a config file from within a session.

### Description
Add small file editors to docker-base

### Examples
I know `nano` is small, but as a `vim` user it would be great to also get a small `vim` version in, something like `vim-tiny`.

Downsides: image size goes up, but it should be at most a MB compared to ~700 we have already. I think that's ok.

corneliusroemer:
Resolves #177

Image size increase should be <1MB unless some libraries are added as well (we'll see in CI):

```bash
$ du -h | grep vim
12K     ./etc/vim
4.0K    ./var/lib/vim/addons
8.0K    ./var/lib/vim
12K     ./usr/share/bug/vim-tiny
32K     ./usr/share/doc/vim-common
32K     ./usr/share/doc/vim-tiny
16K     ./usr/share/vim/vim82/doc
24K     ./usr/share/vim/vim82
28K     ./usr/share/vim

$ du -h | grep nano
4.0K    ./usr/share/doc/nano/examples
12K     ./usr/share/doc/nano
36K     ./usr/share/nano/extra
196K    ./usr/share/nano
4.0K    ./nextstrain/.local/share/nano
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -317,10 +317,12 @@ SHELL ["/bin/bash", "-e", "-u", "-o", "pipefail", "-c"]
 # less: for usability in an interactive prompt
 # libgomp1: for running FastTree
 # libsqlite3: for pyfastx (for Augur)
+# nano: for usability in an interactive prompt
 # ncbi-blast+: for GenoFLU in avian-flu
 # perl: for running VCFtools
 # ruby: may be used by workflows
 # sqlite3: for `augur merge`
+# vim-tiny: for usability in an interactive prompt
 # wget: may be used by workflows
 # zlib1g: for pyfastx (for Augur)
 # nodejs: for running Auspice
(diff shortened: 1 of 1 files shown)


tsibley:
Size difference is about 2MB.

```
$ docker image inspect --format '{{.Size}}' nextstrain/base:branch-add-nano-and-vim-tiny 
2056610544

$ docker image inspect --format '{{.Size}}' nextstrain/base:build-20230905T192825Z
2054525750

$ python
>>> (2056610544 - 2054525750) / 1000 / 1000
2.084794
```

tsibley (approved):
This seems fine (and the size bump is noise given what the image's bloated to), although I'd think the vast majority of folks would edit outside of `nextstrain shell` instead of using `vim` or `nano` inside of it.

Are you thinking we'd make a corresponding change to the Conda runtime too, or not?

genehack:
While reviewing old issues/PRs, I noticed this was approved but still open. It had a couple conflicts with the `Dockerfile` in `master`, so I rebased it to clear those up.

Any reason to not go ahead and merge this?

PR #192, opened 2023-11-06
Include Make in builder-target-platform image

tsibley:
Required for building the Python isal package (for Augur) when there aren't wheels available (such as with recent version 1.5.2).

## Checklist

- [x] Checks pass

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -245,10 +245,12 @@ ARG TARGETARCH
 # git: for git pip installs
 # gcc: for building datrie (for Snakemake)
 # libsqlite3-dev, zlib1g-dev: for building pyfastx (for Augur)
+# make: for building isal (if necessary, for Augur)
 RUN apt-get update && apt-get install -y --no-install-recommends \
         curl \
         gcc \
         git \
+        make \
         jq \
         libsqlite3-dev \
         zlib1g-dev


tsibley:
Merging since this resolves build failures.

Dockerfile
victorlin, line 1: Thanks for looking into this! Where is this isal dependency even from? I don't get it when installing Augur with pip locally.

This fragility is unfortunate, and I wouldn't be surprised if other indirect dependencies cause breakage like this in the future. Would it be worthwhile to provide a standalone installer for Augur for cases like this, where flexibility in dependency versions defined by setup.py is not necessary and even undesirable?

joverlee521: It's a [requirement of xopen for Linux systems only](https://github.com/pycompression/xopen#v110-2021-01-20). (I was curious so I tracked it down with pipdeptree in a nextstrain shell).

tsibley: I don't think a standalone installer for Augur is necessary for this Docker image. It'd be a bit of work for too narrow of a gain, I think, and there are other ways about it. For example, we could take additional measures in the Dockerfile to reduce this sort of breakage by using `--prefer-binary` to the `pip install` of Augur. That would have gotten us the isal 1.5.1 wheels instead of trying to build isal 1.5.2 from source. We could also maintain a constraints file and pass that in with `--constraint` if we wanted to pin things more tightly for this Docker image.

victorlin: @tsibley that makes sense. Good to know those alternatives!

PR #203, opened 2024-01-16
Symlink default nextclade to nextclade3

corneliusroemer:
This might break programs that use old nextclade2 syntax

I've gone through nextstrain workflows to make sure we explicitly use nextclade2/3 so few failures are expected to happen

## Checklist

- [ ] Checks pass


victorlin:
Didn't see this earlier - I've included these changes in #204.

PR #204, opened 2024-01-16
Use Nextclade v3

victorlin:

Unpin Nextclade v3 alpha and use the latest version of Nextclade as default.

Any unmigrated workflows should use `nextclade2` instead of `nextclade`. See the Nextclade v2 → v3 [migration docs](https://docs.nextstrain.org/projects/nextclade/page/user/migration-v3.html) for details.

## Checklist

- [x] Checks pass

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -184,12 +184,8 @@ RUN curl -fsSL https://github.com/lh3/minimap2/releases/download/v2.24/minimap2-
 ARG CACHE_DATE
 
 # Download Nextclade v3
-# Note: Before v3 there used to be separate binaries for Nextclade and
-# Nextalign. Since v3 there is only Nextclade, which can do both.
-# TODO: After Nextclade 3 is released, update the URL to download the latest
-# version instead of hardcoded.
 # TODO: At some point, update the v2 binary symlinks to use Nextclade v3.
-RUN curl -fsSL https://github.com/nextstrain/nextclade/releases/download/3.0.0-alpha.2/nextclade-$(/builder-scripts/target-triple) \
+RUN curl -fsSL https://github.com/nextstrain/nextclade/releases/latest/download/nextclade-$(/builder-scripts/target-triple) \
   -o /final/bin/nextclade3
 
 # Auspice
--- a/Dockerfile
+++ b/Dockerfile
@@ -141,10 +141,8 @@ RUN curl -fsSL https://github.com/nextstrain/nextclade/releases/download/2.14.0/
   && ln -sv nextalign2 /final/bin/nextalign
 
 # Download Nextclade v2
-# Set default Nextclade version to 2
 RUN curl -fsSL https://github.com/nextstrain/nextclade/releases/download/2.14.0/nextclade-$(/builder-scripts/target-triple) \
-  -o /final/bin/nextclade2 \
-  && ln -sv nextclade2 /final/bin/nextclade
+  -o /final/bin/nextclade2
 
 # Download tsv-utils
 # NOTE: Running this program requires support for emulation on the Docker host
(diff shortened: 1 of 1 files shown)


corneliusroemer (approved):
LGTM though no need to use jq here, you could just use `/releases/latest/download`  instead of querying for the latest version 
see https://docs.github.com/en/repositories/releasing-projects-on-github/linking-to-releases

Dockerfile
```diff
@@ -184,13 +183,11 @@ RUN curl -fsSL https://github.com/lh3/minimap2/releases/download/v2.24/minimap2-
...
-# Note: Before v3 there used to be separate binaries for Nextclade and
-# Nextalign. Since v3 there is only Nextclade, which can do both.
-# TODO: After Nextclade 3 is released, update the URL to download the latest
-# version instead of hardcoded.
-# TODO: At some point, update the v2 binary symlinks to use Nextclade v3.
-RUN curl -fsSL https://github.com/nextstrain/nextclade/releases/download/3.0.0-alpha.2/nextclade-$(/builder-scripts/target-triple) \
-  -o /final/bin/nextclade3
+# Set default Nextclade version to 3
+RUN VERSION=$(curl -fsSL https://api.github.com/repos/nextstrain/nextclade/releases/latest | \
+  jq --raw-output '.tag_name') && curl -fsSL https://github.com/nextstrain/nextclade/releases/download/"$VERSION"/nextclade-$(/builder-scripts/target-triple) \
```
corneliusroemer, line 188:
```suggestion
  jq --raw-output '.tag_name') && curl -fsSL https://github.com/nextstrain/nextclade/releases/latest/download/nextclade-$(/builder-scripts/target-triple) \
```
No need for jq this way

victorlin: Oh yeah, I vaguely remember this... `/releases/download/<tag>` vs. `/releases/latest/download`, confusing 😵‍💫

Trying in 3ee286c182564b2bb7b282ff6d27666d53656f02

victorlin: I guess it makes sense since a tag named `latest` is valid

Dockerfile
```diff
@@ -135,16 +135,12 @@ RUN curl -fsSL https://github.com/iqtree/iqtree2/releases/download/v2.1.2/iqtree
 COPY builder-scripts/ /builder-scripts/
 
 # Download Nextalign v2
-# Set default Nextalign version to 2
 RUN curl -fsSL https://github.com/nextstrain/nextclade/releases/download/2.14.0/nextalign-$(/builder-scripts/target-triple) \
-  -o /final/bin/nextalign2 \
-  && ln -sv nextalign2 /final/bin/nextalign
```
victorlin, line 141: Assuming pathogen-repo-ci runs nextalign on the CI example data, I would expect it to fail on the following which still use Nextalign CLI:

- [ncov](https://github.com/nextstrain/ncov/blob/f79bb86c0ba20f685a877e8fa7fbe803ddcf2d9f/workflow/snakemake_rules/main_workflow.smk#L95)
- [seasonal-flu](https://github.com/nextstrain/seasonal-flu/blob/d149b2de52620b8f6f8b675e0215e3cbe01e79e7/workflow/snakemake_rules/core.smk#L69)

Dockerfile
corneliusroemer: Yeah I got confused as well - good point that this is possibly to work around `latest` being a valid tag, didn't think of this

Dockerfile
corneliusroemer: Good point! The lines should stay then!

victorlin: Shouldn't the workflows be updated to either use `nextalign2` explicitly or replace with v3 Nextclade CLI?

victorlin: Oh nevermind, that's not necessary - runtimes will keep `nextalign` around for backwards compatibility. `nextalign` is unambiguously v2 as latest, unlike `nextclade`.

victorlin: Added the lines back in [force-push](https://github.com/nextstrain/docker-base/compare/61a8af29ffa95a9ae9f4aff7c42cee9217ad0c8f..ee302be584707bc1d391931fb8bb2a68ba87d620).

victorlin:
I'm signing off now so feel free to merge this before I get back to it tomorrow.

corneliusroemer (approved):

PR #214, opened 2024-06-06
Use AWS CLI v2

### LINKED ISSUE #213: awscli `--no-sign-request` option not working
## Context 

I was trying to implement the [conditional use of the `--no-sign-request` flag for the avian flu](https://github.com/nextstrain/avian-flu/pull/49) but was still running into "Unable to locate credentials" error. 

I _think_ this is dependency issue within docker-base as I am able to use the flag successfully in conda-base.

joverlee521:

Update AWS CLI to v2

We were running into dependency issues with awscli v1, which was
installed via Pip.¹

Update to AWS CLI v2 which is _not_ installed via Pip.
Instead of working around the absolute path symlinks created by running
`./aws/install`, we are "just installing the files ourselves"² and
creating the symlink for `/final/bin/aws`.

¹ <https://github.com/nextstrain/docker-base/issues/213>
² <https://github.com/nextstrain/docker-base/pull/214#discussion_r1628880435>

## Related issue(s)

Resolves #213 

## Checklist

- [x] Checks pass

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,6 +29,7 @@ COPY --from=xx / /
 # pkg-config: for building VCFtools; may be used by package managers to build from source
 # nodejs: for installing Auspice
 # clang: for compiling C/C++ projects; may be used by package managers to build from source
+# unzip: for AWS CLI
 RUN apt-get update && apt-get install -y --no-install-recommends \
         autoconf \
         automake \
@@ -38,7 +39,8 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         git \
         make \
         pkg-config \
-        dpkg-dev
+        dpkg-dev \
+        unzip
 
 # Install a specific Node.js version
 # https://github.com/nodesource/distributions/blob/4f746d991e099f9494b12e275abe1aab21f1ddf0/README.md#installation-instructions
@@ -216,6 +218,14 @@ RUN curl -fsSL https://ftp.ncbi.nlm.nih.gov/pub/datasets/command-line/v2/linux-$
 RUN curl -fsSL https://ftp.ncbi.nlm.nih.gov/pub/datasets/command-line/v2/linux-${TARGETARCH}/dataformat \
   -o /final/bin/dataformat
 
+# Install tooling for our AWS Batch builds, which use `aws s3`.
+# Using `xx-info march` to translate amd64 -> x86_64 and arm64 -> aarch64
+# based on https://github.com/docker/buildx/discussions/2001
+RUN XX_MARCH=$(xx-info march) \
+  && curl -fsSL --proto '=https' "https://awscli.amazonaws.com/awscli-exe-linux-${XX_MARCH}.zip" -o "awscliv2.zip" \
+  && unzip awscliv2.zip "aws/dist/*" -d /final/libexec/aws-cli \
+  && ln -srv /final/libexec/aws-cli/aws/dist/aws /final/bin/aws
+
 # ———————————————————————————————————————————————————————————————————— #
 
 # Define a builder stage that runs on the target platform.
@@ -251,9 +261,6 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 
 # Install envdir, which is used by pathogen builds
 RUN pip3 install envdir==1.0.1
-
-# Install tooling for our AWS Batch builds, which use `aws s3`.
-RUN pip3 install awscli==1.18.195
 
 # Install Snakemake and related optional dependencies.
 # Pinned to 7.32.3 for stability (2023-09-09)
@@ -436,7 +443,6 @@ COPY --from=builder-target-platform /usr/local/lib/python3.10/site-packages/ /us
 #   -trs, 15 June 2018
 COPY --from=builder-target-platform \
     /usr/local/bin/augur \
-    /usr/local/bin/aws \
     /usr/local/bin/envdir \
     /usr/local/bin/nextstrain \
     /usr/local/bin/pangolin \
--- a/tests/envdir-url.t
+++ b/tests/envdir-url.t
@@ -24,7 +24,7 @@ Files are populated from NEXTSTRAIN_ENVD_URL.
 
   $ aws s3 cp --quiet env.d.zip "$NEXTSTRAIN_ENVD_URL"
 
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -216,6 +218,14 @@ RUN curl -fsSL https://ftp.ncbi.nlm.nih.gov/pub/datasets/command-line/v2/linux-$
 RUN curl -fsSL https://ftp.ncbi.nlm.nih.gov/pub/datasets/command-line/v2/linux-${TARGETARCH}/dataformat \
   -o /final/bin/dataformat
 
+# Install tooling for our AWS Batch builds, which use `aws s3`.
+# Using `xx-info march` to translate amd64 -> x86_64 and arm64 -> aarch64
+# based on https://github.com/docker/buildx/discussions/2001
+RUN XX_MARCH=$(xx-info march) \
+  && curl "https://awscli.amazonaws.com/awscli-exe-linux-${XX_MARCH}.zip" -o "awscliv2.zip" \
+  && unzip awscliv2.zip \
+  && ./aws/install --install-dir /final/bin/aws-cli --bin-dir /final/bin
```
tsibley: On the face of it, I'd suggest this change

```suggestion
  && ./aws/install --install-dir /final/libexec/aws-cli --bin-dir /final/bin
```

which will place the internal executable libraries in a more conventional place, leaving the `bin` dir for just the user-invoked programs.

However, an additional change is needed to fix the broken symlink issue you faced in https://github.com/nextstrain/docker-base/pull/214/commits/f06e64cb54b8e93d42c4486460a5137f6195d0e3, because the installer (`./aws/install`) creates the symlinks with absolute paths like so:

    ln -s /final/libexec/aws-cli/v2/x.y.z /final/libexec/aws-cli/v2/current
    ln -s /final/libexec/aws-cli/v2/current/bin/aws /final/bin/aws

and those target paths (the first paths) don't exist after copying the symlinks (the second paths) into the final image.

We could play tricks on the installer to get it to do what we want by arranging the cwd and `--install-dir` with a relative path just so, but that feels brittle and unobvious. Instead, I'd suggest just installing the files ourselves:

```suggestion
  && mv ./aws/dist /final/libexec/aws-cli \
  && ln -srv /final/libexec/aws-cli/aws /final/bin/aws
```

Also possible to `unzip` directly into place instead of doing `unzip` + `mv`.

Dockerfile
```diff
@@ -216,6 +218,14 @@ RUN curl -fsSL https://ftp.ncbi.nlm.nih.gov/pub/datasets/command-line/v2/linux-$
 RUN curl -fsSL https://ftp.ncbi.nlm.nih.gov/pub/datasets/command-line/v2/linux-${TARGETARCH}/dataformat \
   -o /final/bin/dataformat
 
+# Install tooling for our AWS Batch builds, which use `aws s3`.
+# Using `xx-info march` to translate amd64 -> x86_64 and arm64 -> aarch64
+# based on https://github.com/docker/buildx/discussions/2001
+RUN XX_MARCH=$(xx-info march) \
+  && curl "https://awscli.amazonaws.com/awscli-exe-linux-${XX_MARCH}.zip" -o "awscliv2.zip" \
```
tsibley, line 225:
```suggestion
  && curl -fsSL --proto '=https' "https://awscli.amazonaws.com/awscli-exe-linux-${XX_MARCH}.zip" -o "awscliv2.zip" \
```

Dockerfile
```diff
@@ -436,7 +443,7 @@ COPY --from=builder-target-platform /usr/local/lib/python3.10/site-packages/ /us
 #   -trs, 15 June 2018
 COPY --from=builder-target-platform \
     /usr/local/bin/augur \
-    /usr/local/bin/aws \
+    /usr/local/bin/aws-cli \
```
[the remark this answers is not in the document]
joverlee521: 🤦‍♀️ thanks, I knew I was just missing something obvious here

Dockerfile
joverlee521: > Instead, I'd suggest just installing the files ourselves

Ah, would we not be worried about the filepaths changing? I guess it's pretty unlikely they would change `./aws/dist`.

tsibley: Seems unlikely to change to me, and also seems less likely to break compared to a carefully constructed just-so invocation of `./aws/install` that tightly depends on the internals of `./aws/install`.

joverlee521: Updated in https://github.com/nextstrain/docker-base/pull/214/commits/ef3577f9e1452325ad4280b64c94082f37079e06

tests/envdir-url.t
```diff
@@ -24,7 +24,7 @@ Files are populated from NEXTSTRAIN_ENVD_URL.
 
   $ aws s3 cp --quiet env.d.zip "$NEXTSTRAIN_ENVD_URL"
 
-  $ docker run --rm -e NEXTSTRAIN_ENVD_URL --env=AWS_{ACCESS_KEY_ID,SECRET_ACCESS_KEY,SESSION_TOKEN} "$IMAGE" \
+  $ docker run --rm -e NEXTSTRAIN_ENVD_URL --env=AWS_{ACCESS_KEY_ID,SECRET_ACCESS_KEY,SESSION_TOKEN,DEFAULT_REGION} "$IMAGE" \
```
tsibley, line 27: > AWS CLI v2 runs into errors if the region is not defined¹

Uh... so does this mean that all our usages of the Docker/AWS Batch/Singularity runtimes which call `aws` are going to have to ensure they have `AWS_DEFAULT_REGION` set? That seems like a unexpected breaking change. Hmm.

joverlee521: If that's an issue we don't want to deal with now, we can also [install AWS CLI v1 using the bundled installer](https://docs.aws.amazon.com/cli/v1/userguide/install-linux.html#install-linux-bundled-no-sudo)

tsibley: Maybe? Or maybe we could provide a default `AWS_DEFAULT_REGION` that's backwards compat with v1? (What does it do if region isn't required?) Or maybe we bite the bullet and fix anything that breaks (although this means breaking other people's uses too).

Maybe we can talk thru this next week? I'm not really sure I have all the info at hand here to help make a decision (and it's Friday afternoon and my brain is fried).

genehack: +1 (albeit from deep on the sidelines) for not making this type of decision on Friday afternoon, particularly one with lovely (distractingly) sunny weather...

joverlee521:
Closing in favor of #215 since we don't want to deal with the fallout now while we are dealing with fallout from switch to OIDC in GH Action workflows. We will definitely want to revisit this later though!

PR #215, opened 2024-06-11
Isolate AWS CLI (v1) into its own Python venv

tsibley:
Otherwise, the awscli → botocore dependency is broken by the subsequent installation of the nextstrain-cli → botocore dependency, which results in --no-sign-request not working as it should.

This approach seemed like the option of least change.  Alternatives to and/or additional improvements to this include a) installing Nextstrain CLI in its own venv instead/as well, b) upgrading to AWS CLI v2 (which is always isolated) or c) updating the nextstrain-cli → botocore dependency to not conflict.

Resolves: <https://github.com/nextstrain/docker-base/issues/213>
Related-to: <https://github.com/nextstrain/docker-base/pull/214>

## Checklist

- [x] Checks pass

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -253,7 +253,9 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 RUN pip3 install envdir==1.0.1
 
 # Install tooling for our AWS Batch builds, which use `aws s3`.
-RUN pip3 install awscli==1.18.195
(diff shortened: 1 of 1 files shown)


joverlee521 (approved):
Thanks you @tsibley! Confirmed that this fixes the `--no-sign-request` issue for me!

PR #220, opened 2024-06-17
entrypoint-aws-batch: Upload .snakemake/metadata/ too

tsibley:
Snakemake stores state information per input/output here and uses it to determine if it needs to re-run rules or not.  It seems akin to the file mtimes which we already take care to preserve on upload/download. Additionally, the metadata recorded is used in Snakemake's report generation and is generally useful for looking at workflow statistics.

Continue to not upload all of .snakemake/ en masse because it can potentially contain files that interfere with local usage and/or are large and unnecessary.

Resolves: <https://github.com/nextstrain/cli/issues/373>
Related-to: <https://github.com/nextstrain/cli/pull/374>

## Checklist

- [x] Checks pass

### DIFF
--- a/entrypoint-aws-batch
+++ b/entrypoint-aws-batch
@@ -40,8 +40,8 @@ case "$NEXTSTRAIN_AWS_BATCH_WORKDIR_URL" in
                 return "$zipstatus"
             fi
         }
-        zip -u "$PWD.zip" -r . --exclude ".snakemake/*"     || succeed-when-nothing-to-update $?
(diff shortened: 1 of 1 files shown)


joverlee521 (approved):
Worked as expected in testing with https://github.com/nextstrain/cli/pull/374#pullrequestreview-2124028008

joverlee521:
Hah, totally forgot that ncov-ingest uses another Docker image. Triggered the ncov-ingest [update-image workflow](https://github.com/nextstrain/ncov-ingest/actions/runs/9601291209)

PR #221, opened 2024-06-20
Add pathogen-embed to the base image

huddlej:

Adds pathogen-embed to base image, so we can use its tools for analysis of reassortment in our influenza builds.

## Related issue(s)

https://github.com/nextstrain/conda-base/pull/71

## Checklist

- [x] Checks pass

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -312,6 +312,7 @@ RUN if [[ "$TARGETPLATFORM" == linux/arm64 ]]; then \
       ; \
     fi
 
+RUN pip3 install pathogen-embed==2.0.0
 
 # 2. Add unpinned programs
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -447,6 +447,9 @@ COPY --from=builder-target-platform \
     /usr/local/bin/nextstrain \
     /usr/local/bin/pangolin \
     /usr/local/bin/pangolearn.smk \
+    /usr/local/bin/pathogen-distance \
+    /usr/local/bin/pathogen-embed \
+    /usr/local/bin/pathogen-cluster \
     /usr/local/bin/scorpio \
     /usr/local/bin/snakemake \
     /usr/local/bin/treetime \


Dockerfile
```diff
@@ -312,6 +312,7 @@ RUN if [[ "$TARGETPLATFORM" == linux/arm64 ]]; then \
       ; \
     fi
 
+RUN pip3 install pathogen-embed==2.0.0
```
victorlin, line 315: _(non-blocking)_

Noting from the [run summary](https://github.com/nextstrain/docker-base/actions/runs/9603060473) that this added 8.5 minutes to the build time:

```
[linux/arm64] RUN pip3 install pathogen-embed==2.0.0     | 496.7s (61.9%)  ████████████████████
[linux/amd64] RUN pip3 install pathogen-embed==2.0.0     | 17.0s (2.1%)    ▋
```

I'm not worried since this dependency is pinned, meaning the cached result will be used most of the time and that the increased build time would only be noticeable during (infrequent) cache misses.

victorlin (approved):

Dockerfile
huddlej: This one change also adds ~57 MB to the image file size... 😕 

I expected an increase, since we need to install scikit-learn, HDBSCAN, and UMAP, but it is kind of a bummer.

huddlej:
Even though the CI passed, this PR doesn't currently work as expected. When I download the image for this branch and run a Nextstrain shell like so, the image is missing the executables for `pathogen-distance`, `pathogen-embed`, etc.:

```
$ nextstrain shell --docker --image docker.io/nextstrain/base:branch-add-pathogen-embed .
~/build $ which pathogen-distance
~/build $ which pathogen-embed
~/build $ 
```

However, the Python package is installed, since I can load the Python REPL from the Nextstrain shell and import it:

```
~/build $ python
>>> import pathogen_embed
>>>
```

huddlej:
I forgot this step:

https://github.com/nextstrain/docker-base/blob/3f15ce30e9e9d7e7785e33be179a265e4f69cddd/Dockerfile#L433-L452

Fixing in the next commit.

huddlej:
I confirmed that copying the scripts in the last commit properly provided each of the pathogen-embed tools. I additionally tested that these tools worked in this image with pathogen-embed's cram tests (from the pathogen-embed repo) like so:

```bash
$ nextstrain shell --docker --image docker.io/nextstrain/base:branch-add-pathogen-embed .
~/build $ python -m pip install cram
~/build $ ~/.local/bin/cram --shell=/bin/bash tests
.............................
# Ran 29 tests, 0 skipped, 0 failed.
~/build $
```

I'll plan to merge this PR and the sibling conda-base PR on Monday.

Issue #226, opened 2024-07-25
Add IQtree version with cmaple feature


corneliusroemer:
### Context
I've enabled the cmaple feature in bioconda's iqtree build.

Now what's stopping us from using it in nextstrain builds that are run in ci is lack of that feature in docker-base iqtree.

Binaries are available here: https://github.com/iqtree/iqtree2/releases/tag/v2.3.5.cmaple

Or one could just repackage the bioconda binaries directly - which would have the advantage of `linux-aarch64` being available as well. This would solve #212

corneliusroemer:
As I've mentioned in our Nextstrain zoom call: docker-base is what's holding us back from using iqtree's cmaple mode in workflows with CI. Cmaple mode means we get: a) better trees, b) faster, c) don't need to run a somewhat buggy fix-tree script.

Pinging a few who could be interested/helping here: @huddlej @victorlin @tsibley

tsibley:
The change required here seems trivial? Trying it out in #228.

(Separately, `iqtree2` calling the option that enables CMAPLE `--pathogen` is kinda weird, yeah?)

tsibley:
#228 seems ok to me.

PR #228, opened 2024-07-29
Upgrade IQ-TREE to 2.3.5 with CMAPLE support

tsibley:
We want the CMAPLE support (first available with 2.3.4) for better and faster trees.

CMAPLE support is optional and not all release builds of IQ-TREE have it enabled.  So far, it seems that the "main" releases are made without CMAPLE support (e.g. 2.3.5) and then parallel releases are made with it (e.g. 2.3.5.cmaple), but only for x86_64 on Linux and macOS.  Thus, we're still not using an aarch64/arm64 binary even though the "main" releases (without CMAPLE support) provide them.  If we keep waiting, maybe they'll start being provided, or, eventually, we can start compiling them ourselves.  That's for another time, though.

Resolves: <https://github.com/nextstrain/docker-base/issues/226>

## Checklist

- [x] Checks pass

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -127,7 +127,7 @@ RUN curl -fsSL https://mafft.cbrc.jp/alignment/software/mafft-7.475-linux.tgz \
 # if the processor architecture is not amd64.
 # TODO: Build from source to avoid emulation. Instructions: http://www.iqtree.org/doc/Compilation-Guide
 WORKDIR /download/IQ-TREE
-RUN curl -fsSL https://github.com/iqtree/iqtree2/releases/download/v2.1.2/iqtree-2.1.2-Linux.tar.gz \
+RUN curl -fsSL https://github.com/iqtree/iqtree2/releases/download/v2.3.5.cmaple/iqtree-2.3.5.cmaple-Linux-intel.tar.gz \
   | tar xzvpf - --no-same-owner --strip-components=1 \
  && mv bin/iqtree2 /final/bin/iqtree
 


tsibley:
This upgrades IQ-TREE from 2.1.2 to 2.3.5.cmaple. I scanned the release notes between those versions and didn't notice any major breaking changes.

joverlee521 (commented):
12 ✅ for the pathogen-repo-ci jobs 🎉 

Slightly tangential, does this warrant updating the [citation in augur tree](https://github.com/nextstrain/augur/blob/f814e8634ac3e530196c687bf9d9b988a176d25a/augur/tree.py#L274-L276)?

tsibley:
Uh… maybe!

tsibley:
> 12 ✅ for the pathogen-repo-ci jobs 🎉

\o/

That said, my understanding here is that we'd not necessarily expect these to fail with CMAPLE support but get faster and/or better trees. And our pathogen CI isn't checking tree quality—that's still the sole domain of 👀 unfortunately—so potentially we've accidentally made the trees _worse_ in some way with this change and wouldn't know via automated checks. But that would be contrary to expectations.

huddlej:
Thanks, @tsibley! I'm sure @bqminh will be happy to see this. [He has previous said they would try to prioritize CMAPLE + ARM support](https://github.com/nextstrain/augur/issues/1233#issuecomment-2161809330).

> That said, my understanding here is that we'd not necessarily expect these to fail with CMAPLE support but get faster and/or better trees.

We'd have to opt into the CMAPLE algorithm in our tree rules with the tree builder arg `--pathogen-force` (thanks to @corneliusroemer [for confirming this](https://github.com/nextstrain/augur/issues/1233#issuecomment-2250088829)!). It seems like mpox might be the first place we choose to opt for that algorithm but 🤷🏻

bqminh:
Yes, @trongnhanuit and me are happy to see it happening ;-) thanks!

PR #232, opened 2024-09-04
Add the SQLite CLI for `augur merge`

tsibley:
Ideally this would have been added before or as Augur 25.3.0—the first version to include `augur merge`—was packaged into this runtime, but I plain forgot.

## Checklist

- [x] Checks pass

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -366,6 +366,7 @@ SHELL ["/bin/bash", "-e", "-u", "-o", "pipefail", "-c"]
 # libsqlite3: for pyfastx (for Augur)
 # perl: for running VCFtools
 # ruby: may be used by workflows
+# sqlite3: for `augur merge`
 # wget: may be used by workflows
 # zlib1g: for pyfastx (for Augur)
 # nodejs: for running Auspice
@@ -382,6 +383,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         libsqlite3-0 \
         perl \
         ruby \
+        sqlite3 \
         util-linux \
         wget \
         xz-utils \


tsibley:
Oh, ugh. Our images are based on Debian Bullseye (currently "oldstable"), which [packages SQLite 3.34](https://packages.debian.org/bullseye/sqlite3), but `augur merge` requires ≥3.39. Phooey.

I can either

1. Upgrade image base to Debian Bookworm (currently "stable"), which [packages 3.40](https://packages.debian.org/bookworm/sqlite3). We'll eventually want to do this anyway, but it's a much more wide reaching change and potentially not trivial.
2. Install an appropriate SQLite version from elsewhere, either binaries or compiled from source.

tsibley:
Throwing https://github.com/nextstrain/docker-base/pull/233 at the wall to see what sticks.

tsibley:
Option 1 probably looks something like this:

```diff
diff --git a/Dockerfile b/Dockerfile
index e804b8c..0af2117 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -109,6 +109,17 @@ RUN curl -fsSL https://api.github.com/repos/vcftools/vcftools/tarball/1cab5204eb
  && cp -rp built/bin/*    /final/bin \
  && cp -rp built/share/*  /final/share
 
+# Build SQLite CLI
+WORKDIR /build/sqlite
+RUN curl -fsSL --proto 'https=' https://sqlite.org/2024/sqlite-autoconf-3460100.tar.gz \
+  | tar xzvpf - --no-same-owner --strip-components=1 \
+ && CFLAGS="-Os" ./configure --prefix=$PWD/built \
+      --build=$(TARGETPLATFORM= xx-clang --print-target-triple) \
+      --host=$(xx-clang --print-target-triple) \
+ && make && make install \
+ && cp -rp built/bin/*    /final/bin \
+ && cp -rp built/share/*  /final/share
+
 
 # 2. Download pre-built programs
 
```

tsibley:
Option 1 in https://github.com/nextstrain/docker-base/pull/233 is going to work, it seems, without too much additional effort (and it's effort we'd want/need to do eventually anyway).

PR #233, opened 2024-09-04
Upgrade base image from Debian Bullseye (oldstable) to Bookworm (stable)

tsibley:
Avail ourselves of newer (though still not _new_ ;) software, in particular the SQLite CLI.

## Checklist

- [x] Checks pass

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ FROM --platform=$BUILDPLATFORM tonistiigi/xx AS xx
 # Define a builder stage that runs on the build platform.
 # Even if the target platform is different, instructions will run natively for
 # faster compilation.
-FROM --platform=$BUILDPLATFORM debian:11-slim AS builder-build-platform
+FROM --platform=$BUILDPLATFORM debian:bookworm-slim AS builder-build-platform
 
 SHELL ["/bin/bash", "-e", "-u", "-o", "pipefail", "-c"]
 
@@ -26,7 +26,6 @@ COPY --from=xx / /
 # curl: for downloading source files
 # git: used in builder-scripts/download-repo
 # make: used for building from Makefiles (search for usage); may be used by package managers to build from source
-# pkg-config: for building VCFtools; may be used by package managers to build from source
 # nodejs: for installing Auspice
 # clang: for compiling C/C++ projects; may be used by package managers to build from source
 RUN apt-get update && apt-get install -y --no-install-recommends \
@@ -37,7 +36,6 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         curl \
         git \
         make \
-        pkg-config \
         dpkg-dev
 
 # Install a specific Node.js version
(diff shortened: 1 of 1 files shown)


tsibley:
I expect this to fail in many ways, but let's see what happens!

tsibley:
CI error was during build of vcftools for the aarch64 image. [Relevant part](https://github.com/nextstrain/docker-base/actions/runs/10708388201/job/29690583804?pr=233#step:9:12670) looks like:

```
#35 3.637 checking for aarch64-linux-gnu-pkg-config... no
#35 3.637 checking for pkg-config... /usr/bin/pkg-config
#35 3.638 checking pkg-config is at least version 0.9.0... yes
#35 3.639 checking for zlib... no
#35 3.646 configure: error: Package requirements (zlib) were not met:
#35 3.646 
#35 3.646 Package 'zlib', required by 'virtual:world', not found
#35 3.646 
#35 3.646 Consider adjusting the PKG_CONFIG_PATH environment variable if you
#35 3.646 installed software in a non-standard prefix.
#35 3.646 
#35 3.646 Alternatively, you may set the environment variables ZLIB_CFLAGS
#35 3.646 and ZLIB_LIBS to avoid the need to call pkg-config.
#35 3.646 See the pkg-config man page for more details.
```

We install `zlib1g-dev`, which pulls in `zlib1g`, with `xx-apt-get`. Is the problem that the target platform's `aarch64-linux-gnu-pkg-config` isn't being found and so the build platform's `pkg-config` is used, but zlib isn't installed for the build platform?

genehack (approved):

tsibley:
> Is the problem that the target platform's `aarch64-linux-gnu-pkg-config` isn't being found and so the build platform's `pkg-config` is used, but zlib isn't installed for the build platform?

Replicated the error locally and then validated that installing `pkg-config` with `xx-apt-get` prior to the vcftools build worked.

Dockerfile
```diff
@@ -63,6 +63,7 @@ RUN xx-apt-get install -y \
   gcc \
   g++ \
   libc6-dev \
+  pkg-config \
```
victorlin, line 66: Can it be removed from the `apt-get install` command earlier in this file?

tsibley: Not sure. I didn't look into it. I will!

tsibley:
The [next CI failure](https://github.com/nextstrain/docker-base/actions/runs/10708790365/job/29691923537?pr=233#step:9:28593)

```
> [linux/arm64 final  4/23] RUN if [[ "linux/arm64" == linux/arm64 ]]; then       apt-get update && apt-get install -y --no-install-recommends           libopenblas-base       ;     fi:
0.707 Hit:1 http://deb.debian.org/debian bookworm InRelease
0.709 Hit:2 http://deb.debian.org/debian bookworm-updates InRelease
0.709 Hit:3 http://deb.debian.org/debian-security bookworm-security InRelease
0.793 Hit:4 https://deb.nodesource.com/node_18.x nodistro InRelease

13.16 Reading state information...
13.19 E: Unable to locate package libopenblas-base
```

is because `libopenblas-base` was only a transitional package in Bullseye subsequently removed in Bookworm in favor of the new name, `libopenblas0`. We probably should have used `libopenblas0` back then instead of the transitional package. [Fixed in repush](https://tsibley.net/paste/2024-09-04Lt8eDMdy-rebase.html).

tsibley:
The image built completely in CI but one minor testing failure because I'd based this branch on a month-old `master` (only visible on the workflow `push` event, not the workflow `pull_request` event). Whoops. Rebased to fix, but this is ready for review.

Dockerfile
tsibley: Local testing with a cross-compiled platform says yes, it's not needed earlier. This makes sense as we shouldn't be compiling anything for the build platform, only the target platform, and so should only be using pkg-config for the target platform.

tsibley: [Removed in repush](https://tsibley.net/paste/2024-09-05Mn2_vPSi-rebase.html).

victorlin (approved):
I didn't test anything but changes look good.

tsibley:
The ncov-ingest image seems to build fine locally if I rebase it onto this PR's image.

PR #241, opened 2025-02-12
entrypoint-aws-batch: Keep ../ path parts in ZIP archive members during extraction

tsibley:
entrypoint-aws-batch: Keep ../ path parts in ZIP archive members during extraction

The default of stripping ../ parts in member paths is a (good!) restriction for safety and security, but such paths do not pose any (additional) risk in the context of our Nextstrain runtime containers. We're already downloading and executing arbitrary user-supplied code, so the ability to potentially overwrite system files with ZIP archive members is not any additional privilege. And it's only potential at that due to most files being owned by root in the image, not the default container user of nextstrain.

Keeping the ../ parts will allow Nextstrain CLI to construct ZIP archives for jobs which write to new sibling paths of /nextstrain/build in the container. This will be used for including pathogen workflow source separate (e.g. in /nextstrain/pathogen) from the analysis working directory (/nextstrain/build). It can also be used to support Nextstrain CLI's existing --augur, --auspice, etc. overlays on AWS Batch, though a few other changes are required for that too (coming soon).

Note that Nextstrain CLI does *not* permit ../ path parts when extracting from these same ZIP archives (e.g. after a job completes to download results), as that *would* be additional risk. Currently it strips ../ parts, like unzip's default behaviour, but that will change soon to entirely skip archive members containing ../ parts.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -399,9 +399,17 @@ COPY --from=builder-build-platform /usr/lib/node_modules/ /usr/lib/node_modules/
 # correctly discovered by node.
 RUN ln -sv /usr/lib/node_modules/auspice/auspice.js /usr/local/bin/auspice
 
+# Setup a non-root user for optional use
+RUN useradd nextstrain \
+    --system \
+    --user-group \
+    --shell /bin/bash \
+    --home-dir /nextstrain \
+    --no-log-init
+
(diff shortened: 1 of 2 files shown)


jameshadfield (approved):
Didn't test, but the concept seems sensible!

PR #245, opened 2025-03-13
AWS Batch overlays

tsibley:
See commit messages.

## Checklist

- [x] Checks pass


tsibley:
I'm going to change/extend the approach here a little more to make it independent/more robust to the initial working directory (workdir) not being `/nextstrain/build`.

PR #259, opened 2025-07-18
entrypoint-aws-batch: Upload .snakemake/storage/ too 

joverlee521:

When using Snakemakes storage support to download remote files, Snakemake stores 
the downloaded files in `.snakemake/storage` by default. This allows users to 
keep the remote files that were downloaded during the build on AWS Batch.

Note that the storage path is configurable via Snakemake's 
`--local-storage-prefix`.¹ So if someone configures the storage path to a custom 
path within `.snakemake`, e.g. `.snakemake/foo`, then it would not be available 
in their workdir. I'm not sure it's even possible to make this path configurable 
here, so I'm just going to warn against using it from the Nextstrain CLI. 

Related to https://github.com/nextstrain/cli/pull/460

¹ <https://github.com/snakemake/snakemake/blob/v9.6.3/src/snakemake/cli.py#L1456-L1462>

## Checklist

- [x] Checks pass

### DIFF
--- a/entrypoint-aws-batch
+++ b/entrypoint-aws-batch
@@ -52,8 +52,8 @@ case "$NEXTSTRAIN_AWS_BATCH_WORKDIR_URL" in
                 return "$zipstatus"
             fi
         }
-        zip -u "$PWD.zip" -r . --exclude ".snakemake/*"                             || succeed-when-nothing-to-update $?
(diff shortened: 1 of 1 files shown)


joverlee521:
Worked as expected in https://github.com/nextstrain/cli/pull/460#issuecomment-3098506464

Issue #216, opened 2024-06-12
Revisit updating AWS CLI to v2


joverlee521:
Revisit https://github.com/nextstrain/docker-base/pull/214 once we have the stomach for dealing with fallout of the new requirement for `AWS_DEFAULT_REGION` in AWS CLI v2.

jameshadfield:
Totally understand this is a pain to add, but I've [run](https://github.com/nextstrain/conda/issues/14#issuecomment-947074698) into [bugs](https://github.com/nextstrain/ncov/pull/668#discussion_r656924134) and suggested we do this a few times, and I continue to think it'd be wise to make it an explicit variable (although [not a secret!](https://github.com/nextstrain/rsv/issues/53#issuecomment-1913717467)) throughout our workflows, even in the absence of this issue. Given the requirements of AWS CLI v2 it seems like we should start adding these to workflows / envdirs / etc whenever we notice it's not set.

PR #263, opened 2025-08-19
Allow overriding AWS endpoint url in entrypoint-aws-batch

nik-localstack:

I wanted to test nextstrain agaist [LocalStack](https://www.localstack.cloud/) but the docker container couldn't point to the local s3. I had to make these changes to make it work and I think it may be useful for others too.

## Related issue(s)

This should be handled automatically by newer aws-cli versions , so it kind of relates to https://github.com/nextstrain/docker-base/issues/216

## Checklist

- [x] [Checks](https://github.com/nextstrain/docker-base/actions/runs/17133705517) pass

### DIFF
--- a/tests/aws-batch-endpoint-url.t
+++ b/tests/aws-batch-endpoint-url.t
@@ -1,0 +1,23 @@
+#!/usr/bin/env cram
+
+Setup.
+
+  $ : "${IMAGE:=localhost:5000/nextstrain/base:latest}"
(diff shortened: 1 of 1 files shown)


victorlin (approved):
Thanks! This seems reasonable given that `AWS_ENDPOINT_URL` is the env var name used by AWS CLI v2. I'll push a test and merge after [checks](https://github.com/nextstrain/docker-base/actions/runs/17133705517) pass.

PR #268, opened 2025-09-15
Update output in tests

victorlin:

This started failing as of a CI run last week.¹ I ran locally to reproduce using nextstrain/base:build-20250905T021624Z and applied the diff with cram.

¹ https://github.com/nextstrain/docker-base/actions/runs/17582080191/job/49945576274

## Checklist

- [ ] Checks pass
- [x] ~Update changelog~ N/A


tsibley (approved):
Seems like at 6 months ago the default time style drops the time component for the year.

The other option for addressing this would be to use the `--time-style` option of `ls`, e.g. `--time-style=long-iso` or similar. But I don't think it matters much either way.

PR #290, opened 2026-07-24
Isolate fauna in Python venv

joverlee521:

Fauna is the only blocker on pandas v2 and we don't want to spend the time to upgrading fauna to work with pandas v2, so just isolate it in its own Python venv.

This does require downstream workflows to activate the venv before running any fauna scripts. The only workflow actively using fauna is the seasonal-flu/upload workflow, so this should have minimal impact. See changes in https://github.com/nextstrain/seasonal-flu/pull/341.

## Related issue(s)

Part of https://github.com/nextstrain/public/issues/12

## Checklist

- [x] Checks pass
- [x] ~Update changelog~


victorlin (approved):
Confirmed pandas resolves to v2 with the change.

```sh
$ nextstrain shell --image nextstrain/base:branch-fauna-venv . -c 'python -c "import pandas as pd; print(pd.__version__)"'
...
2.3.3
```

This shouldn't cause any issues downstream, but in case it does, I won't have time to tend to things until Wednesday afternoon after lab meeting presentation.

PR #291, opened 2026-07-27
Node.js v24

jameshadfield:
This PR prepares for Auspice v3, which requires Node.js v24.

I tested the first commit (bumping Node.js to v24) in a local image. This worked well with the latest Auspice release (2.73.0), so can we merge this first commit now ahead of Auspice v3? cc @victorlin 

The second commit (to be dropped - hence the draft PR) shows that Auspice v3 works as expected in the Docker container. There's deprecation warnings due to how `auspice view` should now be invoked, but fixing them requires changes in nextstrain/cli: https://github.com/nextstrain/cli/pull/541


victorlin (approved):
👍 nothing else in the image should depend on Node.js.

Dockerfile
```diff
@@ -216,7 +216,9 @@ RUN curl -fsSL https://github.com/nextstrain/nextclade/releases/latest/download/
 # ⁴ https://github.com/mapbox/node-pre-gyp/blob/v1.0.10/lib/node-pre-gyp.js#L186
 WORKDIR /nextstrain/auspice
 RUN /builder-scripts/download-repo https://github.com/nextstrain/auspice release . \
- && npm install --omit dev && npm link
+ && npm install \
```
victorlin: Minor: the commit message is missing a backtick before `prepare`

Image

victorlin (approved):
I briefly thought about making TypeScript a regular dependency, but realized it's still useful to separate as a dev dependency for the typical `npm install auspice` which doesn't need to run the `prepare` script.

jameshadfield:
> I briefly thought about making TypeScript a regular dependency, but realized it's still useful to separate as a dev dependency for the typical npm install auspice which doesn't need to run the prepare script.

Yeah - the approach our Dockerfile took was somewhat unique: a source install explicitly skipping dev dependencies. I don't know why we didn't install from `npm` but I figure there are good reasons. I've added a sentence to explain this in the commit msg.

victorlin:
The unique approach is to support local [`--auspice`](https://docs.nextstrain.org/projects/cli/en/stable/commands/view/#cmdoption-nextstrain-view-auspice). Confirmed that it still works with v3.

```sh
# make changes to Auspice and rebuild

nextstrain view --image nextstrain/base:branch-auspice-v3 --auspice ~/repos/nextstrain/auspice ~/repos/nextstrain/zika-tutorial
```

